PR #9, opened 2018-10-23
Avoid toggles nil exception

dosuken123:


rarruda:
If you want to add a test for it, that's ok. Otherwise i should add it soon, so we can have it merged.

But I would like to have a test to cover this before merging.

dosuken123:
@rarruda Please add :+1: I can't find time to work on this sadly ... :(

PR #10, opened 2018-10-23
Fix fetching feature values

dosuken123:

### DIFF
--- a/lib/unleash/configuration.rb
+++ b/lib/unleash/configuration.rb
@@ -63,7 +63,7 @@ module Unleash
     end
 
     def fetch_toggles_url
-      self.url + '/client/features'
+      self.url + '/features'
     end
 
     def client_metrics_url
--- a/lib/unleash/toggle_fetcher.rb
+++ b/lib/unleash/toggle_fetcher.rb
@@ -53,7 +53,7 @@ module Unleash
       headers['Content-Type'] = 'application/json'
       headers['If-None-Match'] = self.etag unless self.etag.nil?
 
-      request = Net::HTTP::Get.new(uri.request_uri, headers)
+      request = Net::HTTP::Get.new(uri.request_uri + "?instance_id=#{Unleash.configuration.instance_id}", headers)
 
       response = http.request(request)
 


lib/unleash/configuration.rb
```diff
@@ -63,7 +63,7 @@ def refresh_backup_file!
     end
 
     def fetch_toggles_url
-      self.url + '/client/features'
+      self.url + '/features'
```
ivarconr: Actually the official path for clients to fetch feature toggles are `/client/features`. This was part of the v3 change of Unleash where the "api" and "client" paths was separated in order to allow different authentication mechanisms for clients and the admin ui. Support for the "/features" path directly on root is deprecated in `unleash-server` version 3.x and support for it will disappear in next version. (The server will always be backward compatible one version for clients). 

All version 3.x clients is supposed to fetch toggles via `/client/features`.

lib/unleash/toggle_fetcher.rb
```diff
@@ -53,7 +53,7 @@ def fetch
       headers['Content-Type'] = 'application/json'
       headers['If-None-Match'] = self.etag unless self.etag.nil?
 
-      request = Net::HTTP::Get.new(uri.request_uri, headers)
+      request = Net::HTTP::Get.new(uri.request_uri + "?instance_id=#{Unleash.configuration.instance_id}", headers)
```
ivarconr: These should be added as headers and not as query params. Proper headers to add to all unleash requests are: 
- `UNLEASH-INSTANCEID`
- `UNLEASH-APPNAME`

rarruda:
This PR breaks with the unleash protocol used across the different clients, as explained by @ivarconr 

I would recommend to fix it in gitlab's server implementation instead.

lib/unleash/configuration.rb
dosuken123: @ivarconr Thank you for the explanation. We should fix GitLab side https://gitlab.com/gitlab-org/gitlab-ee/blob/master/ee/lib/api/unleash.rb#L22 indeed. 

Only my concern is that developer who was in charge of this feature tested with golang client, and there were no problems. So that golang client app might be still using deprecated `/features` path

lib/unleash/toggle_fetcher.rb
dosuken123: @ivarconr So the proper way to set client id is

```
# config/initializers/unleash.rb

Unleash.configure do |config|
  config.url      = 'http://unleash.herokuapp.com/api'
  config.app_name = Rails.application.class.parent.to_s
  config.custom_http_headers = {
      UNLEASH-INSTANCEID: 'xxxxxxxx', # Set instance id
      UNLEASH-APPNAME: 'xxxxxxxx', # Set App name 
  }
end

UNLEASH = Unleash::Client.new
```

?

If so, does it make sense to document it on README.md as users can set instance id/app name in this way?

lib/unleash/configuration.rb
dosuken123: @ivarconr The Gitlab's implementation will be fixed in https://gitlab.com/gitlab-org/gitlab-ee/merge_requests/8045

ivarconr: You are right, the go-client is indeed using the deprecated path. I raised an issue to solve this: https://github.com/Unleash/unleash-client-go/issues/30

ivarconr: fyi: It is solved on the v3 branch: https://github.com/Unleash/unleash-client-go/blob/v3/repository.go#L66

lib/unleash/toggle_fetcher.rb
ivarconr: yes, and yes it would of course make sense to document this better.

rarruda: I will fix this in a separate branch/PR. I was not aware of these headers.

It should not be necessary afterwards to manually configure those as `custom_http_headers`.

dosuken123: @rarruda Yes, feel free to ignore this PR. I have already confirmed that the above config can work with GitLab.

PR #11, opened 2018-10-23
Force fetch

dosuken123:

### DIFF
--- a/lib/unleash/client.rb
+++ b/lib/unleash/client.rb
@@ -31,6 +31,7 @@ module Unleash
     def is_enabled?(feature, context = nil, default_value = false)
         Unleash.logger.debug "Unleash::Client.is_enabled? feature: #{feature} with context #{context}"
 
+        Unleash.toggle_fetcher.fetch # Force fetch
         toggle_as_hash = Unleash.toggles.select{ |toggle| toggle['name'] == feature }.first
 
         if toggle_as_hash.nil?


lib/unleash/client.rb
```diff
@@ -31,6 +31,7 @@ def initialize(*opts)
     def is_enabled?(feature, context = nil, default_value = false)
         Unleash.logger.debug "Unleash::Client.is_enabled? feature: #{feature} with context #{context}"
 
+        Unleash.toggle_fetcher.fetch # Force fetch
```
ivarconr: To me it looks like a hack to force a fetch every time one checks the status of a toggle. 

Could you elaborate the need for this one?

dosuken123: @ivarconr Thank you for reviewing. I was just debugging GitLab's new Unleash integration and found those bugs. Please see more details https://gitlab.com/gitlab-org/gitlab-ce/issues/53031

ATM i have no idea how client and server side re-sync again though, currently client side seems not have polling logic to fetch the latest data from server side.

rarruda: The pooling logic lives in https://github.com/Unleash/unleash-client-ruby/blob/master/lib/unleash/toggle_fetcher.rb#L28-L29

ScheduledExecutor (https://github.com/Unleash/unleash-client-ruby/blob/master/lib/unleash/scheduled_executor.rb) will create a new thread and periodically pull the values from the server. How often is configured in the `refresh_interval` config setting.

dosuken123: Ah, thanks :)

It seems the default interval is 15 sec. I'm not sure why it didn't happen even if I waited for a minute, but maybe It could hit `retry_limit`.

ivarconr: It might be that you want guarantees for the initial fetch, so that the client will at least have all toggles from the server before initialisation completes,  this feature was introduced to the java-client a while back (see https://github.com/Unleash/unleash-client-java/pull/58/files)

PR #15, opened 2019-01-03
Initial implementation of a CLI for the unleash client

rarruda:
Useful to test manually what is the status of a FeatureToggle.


README.md
```diff
@@ -171,6 +171,8 @@ end
 ## Local test client
 
 ```
+bundle exec bin/unleash-client
```
rarruda: Add CLI parameters for a simple use case...

PR #16, opened 2019-01-04
Allow the client to shutdown in a clean way

rarruda:


lib/unleash/toggle_fetcher.rb
```diff
@@ -38,7 +35,7 @@ def toggles
     end
 
     # rename to refresh_from_server!  ??
-    # TODO: should simplify by moving uri / http initialization to class initialization
+    # TODO: should simplify by moving uri / http initialization to elsewhere
```
rarruda: typo

lib/unleash/client.rb
```diff
@@ -53,6 +62,27 @@ def is_enabled?(feature, context = nil, default_value = false)
         return toggle_result
     end
 
+    # enabled? is a more ruby idiomatic method name than is_enabled?
+    alias_method :enabled?, :is_enabled?
+
+
```
rarruda: consider a `flush`/`flush_statistics` method, which does not do the `shutdown!` at the end.

PR #18, opened 2019-05-06
ActivationStrategy - missing parameter in default strategy

epicatization:


epicatization:
Hello,
While using default strategy, the response from unleash-server is:
```
{
  "name"=>"test-flag",
  "description"=>"Test flag.",
  "enabled"=>false,
  "strategies"=>[{"name"=>"default"}],
  "variants"=>nil,
  "createdAt"=>"2019-04-17T06:41:19.140Z"
 }
 ```
 So based on 
 https://github.com/Unleash/unleash-client-ruby/blob/4f91fb12eace99ea787050d33b30dc6761d17b8f/lib/unleash/activation_strategy.rb
 It will log warning when `parameters` key is nil.

rarruda:
Thanks for the contribuition/PR. It looks great!

PR #22, opened 2019-09-03
Default result always override result when it set as true

jamesyang124:
After test the result combinations with default result, it does not work as expected when default value is true. It will always override combined result as true regardless the combined result of toggle is enabled and strategy result.

https://github.com/Unleash/unleash-client-ruby/blob/master/lib/unleash/feature_toggle.rb#L74..L81

The modification will be - when strategy result already computed and toggle is enabled, it may respect that combined result instead of default result.

### DIFF
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -72,13 +72,12 @@ module Unleash
 
     # only check if it is enabled, do not do metrics
     def am_enabled?(context, default_result)
-      strategy_result = ( self.strategies.select{ |s|
+      result = self.enabled ? ( self.strategies.select{ |s|
         strategy = Unleash::STRATEGIES.fetch(s.name.to_sym, :unknown)
         r = strategy.is_enabled?(s.params, context)
         Unleash.logger.debug "Strategy #{s.name} returned #{r} with context: #{context}" #"for params #{s.params} "
         r
-      }.any? || self.strategies.empty? )
-      result = self.enabled ? strategy_result : default_result
+      }.any? || self.strategies.empty? ) : default_result
 
       Unleash.logger.debug "FeatureToggle (enabled:#{self.enabled} default_result:#{default_result} and Strategies combined returned #{result})"
       return result
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -72,12 +72,18 @@ module Unleash
 
     # only check if it is enabled, do not do metrics
     def am_enabled?(context, default_result)
-      result = self.enabled ? ( self.strategies.select{ |s|
-        strategy = Unleash::STRATEGIES.fetch(s.name.to_sym, :unknown)
-        r = strategy.is_enabled?(s.params, context)
-        Unleash.logger.debug "Strategy #{s.name} returned #{r} with context: #{context}" #"for params #{s.params} "
-        r
-      }.any? || self.strategies.empty? ) : default_result
+      result =
+        if self.enabled
+          self.strategies.empty? ||
+          self.strategies.select{ |s|
+            strategy = Unleash::STRATEGIES.fetch(s.name.to_sym, :unknown)
+            r = strategy.is_enabled?(s.params, context)
+            Unleash.logger.debug "Strategy #{s.name} returned #{r} with context: #{context}" #"for params #{s.params} "
+            r
+          }.any?
+        else
+          default_result
+        end
 
       Unleash.logger.debug "FeatureToggle (enabled:#{self.enabled} default_result:#{default_result} and Strategies combined returned #{result})"
       return result
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -72,13 +72,13 @@ module Unleash
 
     # only check if it is enabled, do not do metrics
     def am_enabled?(context, default_result)
-      result = self.enabled && ( self.strategies.select{ |s|
+      strategy_result = ( self.strategies.select{ |s|
         strategy = Unleash::STRATEGIES.fetch(s.name.to_sym, :unknown)
         r = strategy.is_enabled?(s.params, context)
         Unleash.logger.debug "Strategy #{s.name} returned #{r} with context: #{context}" #"for params #{s.params} "
         r
       }.any? || self.strategies.empty? )
-      result ||= default_result
+      result = self.enabled ? strategy_result : default_result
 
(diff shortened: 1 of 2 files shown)


lib/unleash/feature_toggle.rb
```diff
@@ -72,13 +72,12 @@ def get_variant(context, fallback_variant = disabled_variant)
 
     # only check if it is enabled, do not do metrics
     def am_enabled?(context, default_result)
```
rarruda: I would like to simplify a bit the `am_enabled?` method, so it is easier to understand.

If you would like, you could re-write it to look a bit more like this?

```ruby
    # only check if it is enabled, do not do metrics
    def am_enabled?(context, default_result)
      result =
        if self.enabled
          self.strategies.empty? ||
          self.strategies.select{ |s|
            r = Unleash::STRATEGIES.fetch(s.name.to_sym, :unknown).is_enabled?(s.params, context)
            Unleash.logger.debug "Strategy #{s.name} returned #{r} with context: #{context}" #"for params #{s.params} "
            r
          }.any?
        else
          default_result
        end

      Unleash.logger.debug "FeatureToggle (enabled:#{self.enabled} default_result:#{default_result} and Strategies combined returned #{result})"
      return result
    end
```

spec/unleash/feature_toggle_spec.rb
```diff
@@ -20,16 +20,126 @@
...
-      name: 'test',
-      enabled: true,
-      strategies: [],
-      variants: nil
+      JSON.parse('{
+          "name": "test",
+          "enabled": true,
+          "strategies": [],
+          "variants": []
+        }')
```
rarruda: While it is possible to `JSON.parse` the string, into the object that the `FeatureToggle` object will use, as a unit test, it would be faster if avoided the deserialization and just sent the hash directly. (as it was done before).

So I don't see the value added in changing this code. For these tests, there is no difference either if `variants` is `nil` or an empty array: `[]`.

spec/unleash/feature_toggle_spec.rb
```diff
@@ -20,16 +20,126 @@
...
     end
+
+    it 'should return true if enabled, and default is false' do
+      context = Unleash::Context.new(user_id: 1)
+      expect(feature_toggle.is_enabled?(context, false)).to be_truthy
+    end
+  end
+
+  describe 'FeatureToggle with empty strategies and disabled toggle' do
+    let(:feature_toggle) { Unleash::FeatureToggle.new(
```
rarruda: Again, I would prefer if the parameters were sent to the constructor as a `Hash` instead of having to go through the extra step of de-serialization.

spec/unleash/feature_toggle_spec.rb
```diff
@@ -20,16 +20,126 @@
...
+
+    it 'should return true if disabled and default is true' do
+      context = Unleash::Context.new(user_id: 1)
+      expect(feature_toggle.is_enabled?(context, true)).to be_truthy
+    end
+  end
+
+  describe 'FeatureToggle with userId strategy and enabled toggle' do
+    let(:feature_toggle) { Unleash::FeatureToggle.new(
+      JSON.parse('{
```
rarruda: Same here! :)

spec/unleash/feature_toggle_spec.rb
```diff
@@ -20,16 +20,126 @@
...
+    end
+
+    it 'should return false if enabled, user_id unmatched, and default is false' do
+      context = Unleash::Context.new(user_id: "54321")
+      expect(feature_toggle.is_enabled?(context, false)).to be_falsey
+    end
+  end
+
+  describe 'FeatureToggle with userId strategy and disabled toggle' do
+    let(:feature_toggle) { Unleash::FeatureToggle.new(
```
rarruda: And here. :)

lib/unleash/feature_toggle.rb
jamesyang124: sure, will patch it later.

spec/unleash/feature_toggle_spec.rb
jamesyang124: I test it with `pry` found the feature toggle object not properly initialized as we configured by the `hash`, so I take JSON and seems like it works. Will verify this later.

jamesyang124:
```ruby
let(:feature_toggle) { Unleash::FeatureToggle.new({
        "name" => "test",
        "enabled" => true,
        "strategies" => [],
        "variants" => nil
      }) }

> feature_toggle
=> #<Unleash::FeatureToggle:0x00007f870b264d70 @enabled=true, @name="test", @strategies=[], @variant_definitions=[]>
```
^ guess the hash's key are string type instead of symbol, so this would not work from original code base:

```ruby
let(:feature_toggle) { Unleash::FeatureToggle.new(
        name: "test",
        enabled: true,
        strategies: [],
        variants: nil
      ) }

> feature_toggle
=> #<Unleash::FeatureToggle:0x00007f94b8965ab0 @enabled=false, @name=nil, @strategies=[], @variant_definitions=[]>
```
and this should support the assumption:

https://github.com/Unleash/unleash-client-ruby/blob/master/lib/unleash/feature_toggle.rb#L14..L21

now will replace JSON impl. as hash with string keys instead.

spec/unleash/feature_toggle_spec.rb
jamesyang124: fix on commit 1ca04ac72fd6edcd507860350b28d74d2172bb9f

spec/unleash/feature_toggle_spec.rb
jamesyang124: fix on commit 1ca04ac72fd6edcd507860350b28d74d2172bb9f

spec/unleash/feature_toggle_spec.rb
jamesyang124: fix on commit 1ca04ac72fd6edcd507860350b28d74d2172bb9f

spec/unleash/feature_toggle_spec.rb
jamesyang124: set back `variants` value as `nil`

fix on commit 1ca04ac72fd6edcd507860350b28d74d2172bb9f

lib/unleash/feature_toggle.rb
jamesyang124: fix on commit d43f0e6f985e9ba801afd3881ab747caee77e148

PR #23, opened 2019-09-05
#20 - Added support for static context fields

rarruda:


ivarconr:
:fire: 
:1st_place_medal:

Issue #31, opened 2020-04-14
Does this client use the retry_limit?


lexicalunit:
I don't see in the code where `retry_limit` is actually used?

rarruda:
Thanks for reporting this bug! This looks like a complete oversight by me. I got a fix in a PR now, I hope I can merge it soon to master.

lexicalunit:
I thought I was just missing it somewhere! Ty for the quick turnaround!

PR #32, opened 2020-04-14
#31 - bugfix: respect retry_limit configuration parameter

rarruda:
configuration setting was never sent to ScheduledExecutor


lexicalunit:
Curious, is there a way to say "make infinite retries" by setting this to 0?

rarruda:
Not as is, but that’s very good feedback.

rarruda:
After looking a bit more into this, i think it would be more elegant to just specify `Float::INFINITY` for explicitly never giving up.

However I see that it can be useful to document this use case, and will add info in the README.

orlandojsjr:
Hey guys, @rarruda are you planning to launch a new release with this PR? ty

rarruda:
@orlandojsjr A new release should be coming out soon!

PR #35, opened 2020-08-09
Fix an issue with undefined default_variant

jejacks0n:
This also addressed a couple instances of duplication around the same working logic.

To have the expected behavior, I had to expose the `disabled_variant` method. I feel like this could use considerably more work getting several aspects that I think might not be tested/working correctly, so if we see value in that I'm happy working at improving the coverage. At the moment though, this at least gets the `get_variant` method working.

### DIFF
--- a/lib/unleash/client.rb
+++ b/lib/unleash/client.rb
@@ -59,24 +59,26 @@ module Unleash
     def get_variant(feature, context = nil, fallback_variant = nil)
       Unleash.logger.debug "Unleash::Client.get_variant for feature: #{feature} with context #{context}"
 
+      default_variant = fallback_variant || Unleash::FeatureToggle.new.disabled_variant
+
       if Unleash.configuration.disable_client
         Unleash.logger.debug "unleash_client is disabled! Always returning #{default_variant} for feature #{feature}!"
-        return fallback_variant || Unleash::FeatureToggle.disabled_variant
+        return default_variant
       end
 
       toggle_as_hash = Unleash&.toggles&.select{ |toggle| toggle['name'] == feature }&.first
 
       if toggle_as_hash.nil?
         Unleash.logger.debug "Unleash::Client.get_variant feature: #{feature} not found"
-        return fallback_variant || Unleash::FeatureToggle.disabled_variant
+        return default_variant
       end
 
       toggle = Unleash::FeatureToggle.new(toggle_as_hash)
-      variant = toggle.get_variant(context, fallback_variant)
+      variant = toggle.get_variant(context, default_variant)
 
       if variant.nil?
         Unleash.logger.debug "Unleash::Client.get_variant variants for feature: #{feature} not found"
-        return fallback_variant || Unleash::FeatureToggle.disabled_variant
+        return default_variant
       end
 
       # TODO: Add to README: name, payload, enabled (bool)
--- a/spec/unleash/client_spec.rb
+++ b/spec/unleash/client_spec.rb
@@ -96,29 +96,31 @@ RSpec.describe Unleash::Client do
 
     features_response_body = '{
       "version": 1,
-      "features": {
-        "name": "toggleName",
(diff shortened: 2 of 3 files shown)


spec/unleash/client_spec.rb
```diff
@@ -96,29 +96,31 @@
...
+                "value": ""
+              }
+            },
+            {
+              "name": "b",
+              "weight": 50,
+              "payload": {
+                "type": "string",
+                "value": ""
+              }
```
jejacks0n, line 119: The existing tests weren't accurate with the actual response the server provides. The server provides an array of features, but the test used to have it returning a hash -- in the code there's places where we iterate over all the features, but this was just iterating over the key/value pairs of the hash and were just coincidentally passing.

spec/unleash/client_spec.rb
```diff
@@ -155,6 +157,13 @@
       unleash_client.is_enabled?('toggleName', {}, true)
     ).to eq(true)
 
+    expect(
+      unleash_client.get_variant('toggleName', Unleash::Context.new(user_id: 42))
+    ).to eq(Unleash::Variant.new(name: 'a', enabled: true, payload: {
+      'type' => 'string',
+      'value' => ''
+    }))
```
jejacks0n, line 165: There appeared to be no tests covering any aspect of the variant logic, so this is the first, and by no means is it comprehensive, but it should generally cover "happy path" at a minimum.

rarruda:
Thanks for the PR and for adding comments to the code in this PR. The comments makes it much easier to understand your line of thought!

I will be looking over the PR soon. Things have been hectic here, but I will try to set aside some time soon.

Nice catch with the wrong data type in the request stub for the variant!

Do you have any specific use cases that the current code is failing? If so, it would be valuable to have tests specifically for that.

lib/unleash/client.rb
```diff
@@ -59,24 +59,26 @@ def if_enabled(feature, context = nil, default_value = false, &blk)
     def get_variant(feature, context = nil, fallback_variant = nil)
       Unleash.logger.debug "Unleash::Client.get_variant for feature: #{feature} with context #{context}"
 
+      default_variant = fallback_variant || Unleash::FeatureToggle.new.disabled_variant
+
       if Unleash.configuration.disable_client
         Unleash.logger.debug "unleash_client is disabled! Always returning #{default_variant} for feature #{feature}!"
```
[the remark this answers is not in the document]
rarruda: I'm thinking that it would be a better approach would be just to just assign `Unleash::FeatureToggle.disabled_variant` in the method signature as the default value instead of nil.

Note that you don't want to instantiate a new object, to get a hold of the disabled variant. It is possible if the disabled_variant method is static. (see comment below in `FeatureToggle.rb`).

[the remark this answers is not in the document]
rarruda: You will also need to rename referenced variable in the log message from `default_variant` to `fallback_variant`.

spec/unleash/client_spec.rb
rarruda: Also worth to note that you only match the Variant "`a`" because of the hashing of the `user_id: 42`. different values in the context could give different results for which variant gets picked.

kazu9su:
Any updates on this? I am still seeing the issue though 😃

jejacks0n:
> Any updates on this? I am still seeing the issue though 😃

Not from me @kazu9su, sorry! I just wanted to fix a bug, and I'm happy if whomever wants to take it over.

The asks in review didn't seem to understand what was being addressed here, or why it's clear that it's missing test coverage, so I didn't feel like making the changes that were asked for just to highlight that . Yes, a _specific_ variant is assigned when one provides a `user: 42`, and _that is in fact the test_.

But I'm happy to provide clarity (if I remember the details), so feel free to include me in any questions if you decide to run with something like this.

PR #43, opened 2021-03-26
Update ruby_version requirement to allow Ruby 3.0

johnny-1111one1:
`rake spec` passed with:

102 examples, 0 failures, 0 pending

PR #45, opened 2021-03-26
Support ruby 3

rarruda:
This PR is a follow up from PR #43


johnny-1111one1:
@rarruda Thanks for the quick turnaround!

PR #56, opened 2021-09-23
Fix backup_file being created too early

mattalat:
`backup_file` is eagerly created when the Configuration is initialized.

 When configuring with a block, the Configuration is initialized with defaults before control is yielded. This causes `backup_file` to be named `unleash--repo.json` regardless of the `app_name` assignment in the block. The filename default should respect block assignment.

### DIFF
--- a/lib/unleash/configuration.rb
+++ b/lib/unleash/configuration.rb
@@ -26,7 +26,7 @@ module Unleash
       initialize_default_logger if opts[:logger].nil?
 
       merge(opts)
-      refresh_backup_file!
+      refresh_backup_file! if app_name.present?
     end
 
     def metrics_interval_in_millis
--- a/spec/unleash/configuration_spec.rb
+++ b/spec/unleash/configuration_spec.rb
@@ -51,6 +51,7 @@ RSpec.describe Unleash do
       expect(Unleash.configuration.url).to eq('http://test-url/')
       expect(Unleash.configuration.app_name).to eq('my-test-app')
       expect(Unleash.configuration.fetch_toggles_url).to eq('http://test-url//client/features')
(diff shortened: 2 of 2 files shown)


mattalat:
@rarruda How do you think I should handle this case while still respecting cases where no `app_name` is present by design? I could recreate or rename the file at the end of `Unleash.configure` before calling the refresh method.

rarruda:
`backup_file` is eagerly configured/set when the Configuration is initialized.

The issue is in:

https://github.com/Unleash/unleash-client-ruby/blob/master/lib/unleash/configuration.rb#L44

As the `refresh_backup_file!` method does not refresh (in `/lib/unleash.rb#38`) given that it was previously configured before the handling of the `yield`  ( https://github.com/Unleash/unleash-client-ruby/blob/master/lib/unleash.rb#L34 )

One possible solution would be to remove the `if` check in the end of line `/lib/unleash/configuration.rb#L44`. 

I would like to write some tests and see if there are any possible side effects on the change. (But feel free to also do that, and share what you find in this PR if you want).

mattalat:
> I would like to write some tests and see if there are any possible side effects on the change. (But feel free to also do that, and share what you find in this PR if you want).

@rarruda I just got back on this but I see you made some changes upstream. Do you still need any help with ☝️?

rarruda:
@mattalat Sorry for the slow answer. This is now fixed in #63 ! Thanks for pointing this out!

I chose a slightly different approach than the one you suggested in the PR, but it was quite useful to have your input. Highly appreciated.

I will try to cut out a new point release with this fix soon.

mattalat:
No worries! Happy to help. Thanks for your work—we're making use of it!

Issue #58, opened 2021-09-29
fix: metrics only sent if not empty


ivarconr:
The Ruby SDK sent metrics even if it is empty. Instead it can hold on to the local metrics longer if there are no metrics to report (e.g. up to 10min).

rarruda:
This feature is now merged in master.

PR #66, opened 2021-11-13
fix: only send metrics if there is data to send. (#58)

rarruda:
We want to preserve the backend from having to work, when there is no work to be done.

### DIFF
--- a/lib/unleash/metrics_reporter.rb
+++ b/lib/unleash/metrics_reporter.rb
@@ -6,6 +6,8 @@ require 'time'
 
 module Unleash
   class MetricsReporter
+    LONGEST_WITHOUT_A_REPORT = 600
+
     attr_accessor :last_time
 
     def initialize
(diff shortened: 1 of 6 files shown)


spec/unleash/metrics_reporter_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+    Unleash.toggle_metrics.increment('featureA', :yes)
+    Unleash.toggle_metrics.increment('featureA', :yes)
+    Unleash.toggle_metrics.increment('featureA', :yes)
+    Unleash.toggle_metrics.increment('featureA', :no)
+    Unleash.toggle_metrics.increment('featureA', :no)
+    Unleash.toggle_metrics.increment('featureB', :yes)
+
+    metrics_reporter.post
+
+    expect(WebMock).to have_requested(:post, 'http://test-url/client/metrics')
```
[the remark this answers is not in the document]
rarruda: @sonatype-lift ignore

ivarconr (approved):

lib/unleash/metrics_reporter.rb
```diff
@@ -33,16 +35,28 @@ def generate_report
...
 
-    def send
-      Unleash.logger.debug "send() Report"
+    def post
+      Unleash.logger.debug "post() Report"
+
+      if bucket_empty? && (Time.now - self.last_time < LONGEST_WITHOUT_A_REPORT) # and last time is less then 10 minutes...
+        Unleash.logger.debug "Report not posted to server, as it would have been empty. (and has been empty for up to 10 min)"
+
+        return
```
sighphyre, line 44: If I'm reading this correctly, the early return here means that self.last_time won't get updated in this call. If that's the case, the logs will get thumped with the warning message until a report actually gets posted. Is that the behavior we want here or is it better to update the time just prior to this so that this warning only gets raised every 10 minutes?

rarruda: That's correct. It's a debug log statement, so it shouldn't be visible for more users anyhow, unless they more or less ask for it.

Given that it is a debug log statement, I don't think we should delay printing it out.

sighphyre (approved):
Looks good

PR #67, opened 2021-11-21
Add Truffleruby head to CI

gogainda:


rarruda:
Hi! Thanks for the suggestion! Would you care to move truffleruby right after jruby in the list (and before the MRI versions)? (as they are both jvm based implementations of Ruby).

Also, I would prefer to not test agains head, but against their regular releases, so instead of `truffleruby-head` could you switch to `truffleruby`?

If you think these are too much hassle, that's ok, just let me know and I can do these in a separate PR.

rarruda:
Also if it all goes well, we can also add truffleruby to the list of supported rubies in the README. :)

rarruda:
Locally it seems like truffleruby (both stable: `20.3.0` and latest) doesn't work as expected: at first glance I get multiple tests failing where MRI and Jruby seem to pass. 

I have not investigated into the causes, but we should ensure that it works as expected in truffleruby before adding it to the supported versions of ruby.

gogainda:
Looks like only head is passing right?

rarruda:
I get many failures on truffleruby-head too...

```sh
$ rvm use truffleruby-head
Using ~/.rvm/gems/truffleruby-head
$ bundle exec rspec --format failures
(...)
./spec/unleash/client_specification_spec.rb:38:test that Feature.B3 should be enabled for userId=122
./spec/unleash/client_specification_spec.rb:38:test that Feature.B4 should be enabled for sessionId=122
./spec/unleash/client_specification_spec.rb:52:test that Feature.Variants.C should return variant1
./spec/unleash/client_specification_spec.rb:52:test that Feature.Variants.C should return variant3
./spec/unleash/client_specification_spec.rb:52:test that Feature.Variants.D should return variant1
./spec/unleash/client_specification_spec.rb:52:test that Feature.Variants.D should return variant2
./spec/unleash/client_specification_spec.rb:52:test that Feature.Variants.D should return variant3
./spec/unleash/client_specification_spec.rb:38:test that Feature.flexibleRollout.10 should be enabled for userId=174
./spec/unleash/client_specification_spec.rb:38:test that Feature.flexibleRollout.10 should be enabled for sessionId=174
(...)
./spec/unleash/feature_toggle_spec.rb:181:should return variant2 for user_id:2
./spec/unleash/feature_toggle_spec.rb:236:should return variantB for user_id:2
(...)
./spec/unleash/strategy/flexible_rollout_spec.rb:9:should always be enabled when rollout is set to 100, disabled when set to 0
./spec/unleash/strategy/util_spec.rb:6:returns correct values
Coverage report generated for RSpec to /home/rarruda/git/unleash/unleash-client-ruby/coverage. 1241 / 1316 LOC (94.3%) covered.
Lcov style coverage report generated for RSpec to coverage/lcov.info
Stopped processing SimpleCov as a previous error not related to SimpleCov has been detected
```

I downloaded truffleruby for the first time yesterday.

gogainda:
could you paste ruby version?

rarruda:
sure!

```
$ rvm use truffleruby-head
Using /home/rarruda/.rvm/gems/truffleruby-head
$ ruby --version
truffleruby 22.0.0-dev-f439b388, like ruby 3.0.2, GraalVM CE Native [x86_64-linux]
```

and 

```
$ rvm use truffleruby-20.3.0
Using /home/rarruda/.rvm/gems/truffleruby-20.3.0
$ ruby --version
truffleruby 20.3.0, like ruby 2.6.6, GraalVM CE Native [x86_64-linux]
```

gogainda:
indeed:
```
Finished in 2.62 seconds (files took 3.73 seconds to load)
113 examples, 4 failures, 2 pending
```

rarruda:
Given that all tests pass fine in jruby and MRI 2.5 to 3.0, I suspect it might be something with truffleruby or a dependency. 

Feel free to investigate. But I can't merge the PR as is.

PR #69, opened 2021-12-01
feat: Implement custom stickiness

sighphyre:
- Add custom stickiness for feature variants
- Add custom stickiness for flexible rollout strategies
- Implement custom stickiness property on variant defs
- Force context to resolve to nil instead of crashing with empty props

### DIFF
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -40,7 +40,7 @@ module Unleash
       return disabled_variant unless self.enabled && am_enabled?(context, true)
       return disabled_variant if sum_variant_defs_weights <= 0
 
-      stickiness = self.variant_definitions[0].stickiness
+      stickiness = self.variant_definitions[0].stickiness || "default"
 
       variant = variant_from_override_match(context)
       variant = variant_from_weights(context, stickiness) if variant.nil?
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -40,8 +40,10 @@ module Unleash
       return disabled_variant unless self.enabled && am_enabled?(context, true)
       return disabled_variant if sum_variant_defs_weights <= 0
 
+      stickiness = self.variant_definitions[0].stickiness
+
       variant = variant_from_override_match(context)
-      variant = variant_from_weights(context) if variant.nil?
+      variant = variant_from_weights(context, stickiness) if variant.nil?
 
       Unleash.toggle_metrics.increment_variant(self.name, variant.name) unless Unleash.configuration.disable_metrics
       variant
@@ -85,7 +87,8 @@ module Unleash
       self.variant_definitions.map(&:weight).reduce(0, :+)
     end
 
-    def variant_salt(context)
+    def variant_salt(context, stickiness = "default")
+      return context.get_by_name(stickiness) unless stickiness == "default"
       return context.user_id unless context.user_id.to_s.empty?
       return context.session_id unless context.session_id.to_s.empty?
       return context.remote_address unless context.remote_address.to_s.empty?
@@ -100,8 +103,10 @@ module Unleash
       Unleash::Variant.new(name: variant.name, enabled: true, payload: variant.payload)
     end
 
-    def variant_from_weights(context)
-      variant_weight = Unleash::Strategy::Util.get_normalized_number(variant_salt(context), self.name, sum_variant_defs_weights)
+    def variant_from_weights(context, stickiness)
+      seed = variant_salt(context, stickiness)
+
+      variant_weight = Unleash::Strategy::Util.get_normalized_number(seed, self.name, sum_variant_defs_weights)
       prev_weights = 0
 
       variant_definition = self.variant_definitions
@@ -150,6 +155,7 @@ module Unleash
             v.fetch('name', ''),
             v.fetch('weight', 0),
             v.fetch('payload', nil),
+            v.fetch('stickiness', nil),
             v.fetch('overrides', [])
           )
         end || []
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -32,6 +32,14 @@ module Unleash
       result
     end
 
+    def resolve_variant_stickiness(variant_definitions)
+      return nil if variant_definitions.nil?
+      sticky_variant = variant_definitions.find { |variant|
+        !variant.nil?
+      }
+      sticky_variant && sticky_variant.stickiness
+    end
+
     def get_variant(context, fallback_variant = disabled_variant)
       raise ArgumentError, "Provided fallback_variant is not of type Unleash::Variant" if fallback_variant.class.name != 'Unleash::Variant'
 
@@ -40,7 +48,7 @@ module Unleash
       return disabled_variant unless self.enabled && am_enabled?(context, true)
       return disabled_variant if sum_variant_defs_weights <= 0
 
-      stickiness = self.variant_definitions[0].stickiness || "default"
+      stickiness = resolve_variant_stickiness(self.variant_definitions) || "default"
 
       variant = variant_from_override_match(context)
       variant = variant_from_weights(context, stickiness) if variant.nil?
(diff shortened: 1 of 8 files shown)


ivarconr:
@sighphyre looks like integration tests for variants is currently failing 
(https://github.com/Unleash/client-specification/blob/master/specifications/08-variants.json)

rarruda:
Will review this as soon as i get a chance. This PR should implement #41 and #42 ! :rocket:

rarruda (commented):
I would like to have #71 merged (and this PR rebased on top of it) before approving it, as it does have a change which contains side-effects.

lib/unleash/context.rb
```diff
@@ -28,7 +28,9 @@ def get_by_name(name)
       if ATTRS.include? normalized_name
         self.send(normalized_name)
       else
-        self.properties.fetch(normalized_name)
+        self.properties.fetch(normalized_name) {
+          || self.properties.fetch(name.to_sym, nil)
+        }
```
rarruda, line 33: This change has side-effects with current behaviour (which is tested explicitly in #71). 

IIRC this change sorta makes sense, (though I would have prefered a different syntax here).

```suggestion
        self.properties.fetch(normalized_name, nil) || self.properties.fetch(name.to_sym, nil)
```

But I would prefer to keep this as a separate PR / commit / etc, if possible.

lib/unleash/feature_toggle.rb
```diff
@@ -40,8 +40,10 @@ def get_variant(context, fallback_variant = disabled_variant)
       return disabled_variant unless self.enabled && am_enabled?(context, true)
       return disabled_variant if sum_variant_defs_weights <= 0
 
+      stickiness = self.variant_definitions[0].stickiness || "default"
```
rarruda, line 43: The `[0]` makes me uncomfortable. 

Why should it take the first variant in the list? What happens if you try to `get_variant()` from a feature toggle with no variants? (`self.variant_definitions[0]` would be `nil` and `nil.stickiness` would throw a `NoMethodError`).

sighphyre: So we've included the stickiness in each variant in the list for backwards compatibility reasons in the variant list

variant_definitions _should_ never be nil but I realize these are famous last words and your comment is entirely fair. Let me take another look at this today and see if I can structure this in a way that doesn't looks so dangerous

lib/unleash/context.rb
sighphyre: Syntax here looks nice, thanks for the suggestion!

lib/unleash/feature_toggle.rb
sighphyre: I want to say this should be safer now - code should try find the first non nil variant or default to "default"

lib/unleash/strategy/flexible_rollout.rb
```diff
@@ -38,16 +38,12 @@ def random
...
-          context.user_id
-        when 'sessionId'
-          context.session_id
         when 'random'
           random
         when 'default'
           context.user_id || context.session_id || random
         else
-          nil
+          context.get_by_name(stickiness)
```
rarruda, line 46: you might need to do a `begin`/`rescue` for `KeyError` here in the meantime, if you want to keep this behavior in place.

rarruda (changes requested):
This is great! But I do have a couple of concerns that I would like to see addressed before merging:

- change the code for resolving stickiness to something more ruby idiomatic
- not changing the signature (results) for `get_by_name()` in this PR.
- rebase this branch to master, and run `bundle exec rubocop` to catch any code style deviations from the project standard.

lib/unleash/context.rb
```diff
@@ -28,7 +28,7 @@ def get_by_name(name)
       if ATTRS.include? normalized_name
         self.send(normalized_name)
       else
-        self.properties.fetch(normalized_name)
+        self.properties.fetch(normalized_name, nil) || self.properties.fetch(name.to_sym, nil)
```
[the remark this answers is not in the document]
sighphyre: Yeah, I think that makes a lot of sense to pull this out into a separate PR. I'll make one and push it shortly. Unfortunately, this won't be able to get merged until that gets merged in. The resolving to a nil over raising an error is secondary to the issue that properties don't resolve correctly on the context object (or they do and I've done something silly!) This felt like the least invasive change to make, given that this is my first PR on the project, but probably there's a cleaner fix and we should discuss that and do the correct thing over the expedient thing

lib/unleash/strategy/flexible_rollout.rb
sighphyre: I think it's probably just worth waiting for us to settle on fixing the context issue first. Resolving the nil is secondary to getting the correct properties back

lib/unleash/feature_toggle.rb
```diff
@@ -40,8 +44,7 @@ def get_variant(context, fallback_variant = disabled_variant)
       return disabled_variant unless self.enabled && am_enabled?(context, true)
       return disabled_variant if sum_variant_defs_weights <= 0
 
-      variant = variant_from_override_match(context)
-      variant = variant_from_weights(context) if variant.nil?
+      variant = variant_from_override_match(context) || variant_from_weights(context, resolve_stickiness)
```
rarruda, line 47: I am not sure if this is a real issue or not, but if the variant is not nil, then variants from weights will be called. (Before it would not). 

But this could maybe theoretically change behavior? But I think tests would have caught it, if it would have been a real issue.

lib/unleash/feature_toggle.rb
```diff
@@ -32,6 +32,10 @@ def is_enabled?(context, default_result)
       result
     end
 
+    def resolve_stickiness
+      self.variant_definitions&.map(&:stickiness)&.compact&.first || "default"
+    end
```
rarruda, line 37: Could you double check that this method is in the private section? (Can't see from my mobile).

lib/unleash/feature_toggle.rb
rarruda: Could be that it just wastes some CPU cycles?

lib/unleash/feature_toggle.rb
sighphyre: Spot on - it's been moved!

lib/unleash/feature_toggle.rb
sighphyre: So I've run through this a million times in my mind and I've written some scratch-pad code to test this for myself. Unless there's something subtle about the way Ruby resolves function calls I don't think this is the case. 

If variant_from_override_match resolves to any truthy value then boolean short circuiting should kick in and variant_from_weights won't be called. 

If variant_from_override_match resolves to a falsey value then the  variant_from_weights will execute.

While this is technically a logic change here, since we're no longer handling just nils and now capturing that whole case with falsely values, I don't think it matters since variant_from_override_match can only return a match, a new value or a nil.

If you consider the following sample code:

class Test
  def fail_hard
    raise StandardError, 'This is broken'
  end

  def return_input(input)
    puts "I was called with #{input}"
    input
  end
end

class Empty
end

test = Test.new
test.return_input(true) || test.fail_hard
test.return_input(Empty.new) || test.fail_hard
test.return_input(nil) || test.fail_hard
`

When run this yields the following output:

I was called with true
I was called with #<Empty:0x00005593a2beda58>
I was called with 
Traceback (most recent call last):
	1: from main.rb:18:in `<main>'
main.rb:3:in `fail_hard': This is broken (StandardError)

I can't see a meaningful difference between the toy example here and the code in question.

...but it's Friday and it's been a long week so I'm open to being told I'm crazy

`

lib/unleash/context.rb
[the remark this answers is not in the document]
sighphyre: Resolved by #74

lib/unleash/feature_toggle.rb
rarruda: This is a non-issue, I just wasn't able to read clearly on friday.

rarruda (approved):
Looks good! I can't spot anything strang anymore :)

PR #71, opened 2021-12-02
tests: make tests pass in context matching

rarruda:
make it explicit the current state of matching.

It would be good to have this PR merged before #69  as this includes tests for some use-cases that are affected there.


sighphyre (approved):
Sure, looks good!

PR #74, opened 2021-12-06
fix: Allow context to correctly resolve camelCase property values

sighphyre:
Fix an issue where camel case property values being resolved in the context object would incorrectly raise a KeyError


ivarconr (approved):
nice =)

PR #78, opened 2021-12-16
chore: standardize default values for metrics_interval and retry_limit

rarruda:
to the same values used in the other SDKs


ivarconr (approved):

chriswk (approved):
Looks good to me

Issue #64, opened 2021-11-09
feat: Add support for boostrap


ivarconr:
Some context: 
https://www.getunleash.io/blog/bootstrap-unleash-java-sdk-to-improve-the-resilience

rarruda:
- Bootstrap from static local file
- Bootstrap from custom provider

rarruda:
This was  implemented in PR #85 which is now merged. Closing.

PR #85, opened 2022-01-25
feat: Implement custom bootstrapping on startup

sighphyre:
Adds the ability to bootstrap Toggle states from a custom source on load of the client. Currently, there's direct support for loading from a URL and/or loading from a file


sighphyre:
This is related to https://github.com/Unleash/unleash-client-ruby/issues/64

This adds support for loading from a file,URL or custom implementation

lib/unleash/bootstrap.rb
```diff
@@ -0,0 +1,55 @@
+require 'unleash/configuration'
+require 'unleash/feature_toggle'
+require 'logger'
+require 'time'
+require 'net/http'
+require 'uri'
+
+module Unleash
```
ivarconr, line 8: It would be nice to also see a `data` bootstrapper out of the box.

lib/unleash/toggle_fetcher.rb
```diff
@@ -4,20 +4,26 @@
...
       self.toggle_cache = nil
       self.toggle_lock = Mutex.new
       self.toggle_resource = ConditionVariable.new
       self.retry_count = 0
+      self.bootstrapper = Unleash.configuration.bootstrapper
 
       # start by fetching synchronously, and failing back to reading the backup file.
       begin
-        fetch
+        if !self.bootstrapper.nil? # if the consumer provides a bootstrapper, we're going to assume they want to use it
```
ivarconr, line 19: I am unsure if I understand this code block. We want to:
1. Load configuration from the boostrapper 
2. Load configuration from the Unleash API (fetch) as soon as possible.

lib/unleash/bootstrap.rb
```diff
@@ -0,0 +1,55 @@
...
+    def build_headers(headers = nil)
+      headers = (headers || {}).dup
+      headers['Content-Type'] = 'application/json'
+
+      headers
+    end
+  end
+
+  def self.extract_bootstrap(bootstrap_hash)
+    raise NotImplemented, "The provided bootstrap doesn't seem to be a valid set of toggles" if bootstrap_hash['version'] < 1
```
ivarconr, line 51: I am unsure if we want to require the version field here.

rarruda (commented):
I would like to see Bootstrap as a separate Module, with different classes in it, each providing an implementation of an abstract class.

Note that ideally each class should be in a separate file.

For example please look at https://github.com/Unleash/unleash-client-ruby/blob/main/lib/unleash/strategy/base.rb and https://github.com/Unleash/unleash-client-ruby/blob/main/lib/unleash/strategy/default.rb

If you want/have the time/patience, I can try creating a branch on top of yours to show what I mean, but It could take me a couple of days until i can get around to it.

lib/unleash/bootstrap.rb
```diff
@@ -0,0 +1,55 @@
...
+  class UrlBootStrapper
+    attr_accessor :uri, :headers
+
+    def initialize(uri, headers)
+      self.uri = URI(uri)
+      self.headers = headers
+    end
+
+    def read
+      request = Net::HTTP::Get.new(self.uri, self.build_headers(self.headers))
```
rarruda, line 32: Note that as a "Built-in" class, then it should use the functionalities from Unleash::Util::Http

lib/unleash/bootstrap.rb
rarruda: So, this is now possible after #86 got merged.

The main concern is that the data would need to be serialized in JSON first (to only then be de-serialized).

Having an extra `Unleash::Bootstrap::FromData` class doesn't seem like a bad idea though.

lib/unleash/bootstrap.rb
rarruda: Fixed after the merge of #86

README.md
```diff
@@ -267,6 +268,27 @@ variant = UNLEASH.get_variant "ColorVariants", @unleash_context, fallback_varian
...
+
+`bootstrap_data` configuration allows the client to be initialized with a predefined set of toggle states.
+The content of the parameter is a JSON string containing the response body from the unleash server.
+
+We provide two classes to help fetch the bootstrap files:
+* `Unleash::Bootstrap::FromFile`
+* `Unleash::Bootstrap::FromUri`
+
+Example usage:
+```ruby
```
rarruda, line 281:
```suggestion

First saving the toggles locally:
```shell
curl -H 'Authorization: abc' -XGET 'http://unleash.herokuapp.com/api' > ./default-toggles.json
```

Now using them on start up:
```ruby
```

README.md
```diff
@@ -305,6 +327,12 @@ This client comes with the all the required strategies out of the box:
  * UnknownStrategy
  * UserWithIdStrategy
 
+## Available Bootstrap Classes
+
+This client comes with these classes to load unleash features on startup, before making a request to the Unleash API:
+
+ * Unleash::Bootstrap::FromFile
+ * Unleash::Bootstrap::FromUri
```
rarruda, line 335: Maybe add examples, including how to set HTTP headers.

examples/default-toggles.json
```diff
@@ -0,0 +1,43 @@
+
```
rarruda, line 1:
```suggestion
```

examples/default-toggles.json
```diff
@@ -0,0 +1,43 @@
...
+            {
+                "name": "hola",
+                "parameters": {
+                    "name": "val"
+                }
+            }
+        ]
+
+    }
+]}
```
rarruda, line 43:
```suggestion
]
}
```

spec/unleash/bootstrap/from_uri_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'unleash/bootstrap/from_uri'
+require 'json'
+
+RSpec.describe Unleash::Bootstrap::FromUri do
+  it 'loads bootstrap toggle correctly from URL' do
+    bootstrap_file = './spec/unleash/bootstrap-resources/features-v1.json'
```
rarruda, line 6: maybe consider in-lining the JSON here. The data is not really re-used in different contexts.

lib/unleash/toggle_fetcher.rb
rarruda: Is it ok to wait for the next update cycle? Or should that happen immediately after loading via the bootstraping mechanism?

lib/unleash/bootstrap.rb
rarruda: In the merged (from #86) we use the same test for the fetched toggles as we do for the bootstrapped ones. Which seems pretty reasonable imho.

README.md
```diff
@@ -267,6 +268,27 @@ variant = UNLEASH.get_variant "ColorVariants", @unleash_context, fallback_varian
...
+#### Bootstrapping
+
+`bootstrap_data` configuration allows the client to be initialized with a predefined set of toggle states.
+The content of the parameter is a JSON string containing the response body from the unleash server.
+
+We provide two classes to help fetch the bootstrap files:
+* `Unleash::Bootstrap::FromFile`
+* `Unleash::Bootstrap::FromUri`
+
+Example usage:
```
rarruda, line 280:
```suggestion
Example usage:

First saving the toggles locally:
```shell
curl -H 'Authorization: abc' -XGET 'http://unleash.herokuapp.com/api' > ./default-toggles.json
```.

Now using them on start up:
```

PR #86, opened 2022-01-29
feat: make bootstrapping more ruby idiomatic

rarruda:
I refactored the code to make it more ruby idiomatic.

But there are a few things that I still am not quite satisfied yet, so I will come back to that in a second commit.

The things that come to mind right now:
- bootstrapper should take a string, with the feature json blob. There is no need to send an entire class our way.

For bonus points it also:
- should consider allowing to use the bootstrapped data, even when the `disable_client` is true.
- tests for what happens when loading of the bootstrapped file/data fails.
- tests when the bootstrap data is set/sent from the client initialization
- handle environment variables `UNLEASH_BOOTSTRAP_FILE` and `UNLEASH_BOOTSTRAP_URL` as it is done in the Java SDK


sighphyre:
Very cool! The tests are awesome

So a few thoughts, first, it sounds like you have concern over passing a class around for the bootstrapper. Mind unpacking that a little? What price do you see us paying by requiring a class here? A class probably isn't needed right now but allowing us to delegate the read method to a later point in time gives flexibility for future work, it also allows us to control what exception types bubble up. As is, this puts handling errors from missing files/closed sockets firmly in the hands of the user without warning them. Ideally, I'd like to catch what we can from any bootstrapping tasks, including loading data sources and warn without crashing the call stack. If we gotta pass in data rather than handling lazily, is there a more appropriate data structure we can use here?

rarruda:
What kind of future work?

I can agree that having a standardized API across SDKs  is really desirable though.

The burden on providing valid data will always fall on the user of the SDK either way. We still bundle helper classes for developers that don't want to implement the interface to fetch the data via File / HTTP.

Another possibility would be to just pass a lambda, but I am not super convinced of that either.

But given that this thread is now in a closed PR, maybe we can follow up in #85 ?

ivarconr:
**tldr;**
1. I do not think it is enough to only accept `bootstrap_data`. 
2. We should do that in combination with giving the user the ability to specify standard locations such as `url` and `file`, so that we are in control of when it is loaded, and open the door for fetching things in parallel. 
3. We need to also allow for some flexibility by allowing the user to take control of the actual loading, but we need to still be in control of the interface and data format, and when and if the execution happens. This can be solved with a lambda or though a interface / class. 

---------------------------
For context the main motivation for adding bootstrap to all SDKs is to make it easy to bootstrap from the Unleash Proxy ([PR](https://github.com/Unleash/unleash-proxy/pull/44)). Thus we need to make it super easy to specify a bootstrap URL and the authorization header required to do that. 

I general I think we need to balance different aspects:

**1. Make the SDK easy to use**
  - This without requiring a lot of extra, potentially manual, steps to do "common" things. 
  - This is why we want to make it super intuitive to load the bootstrap from standard location like `a url`, `a file` or provided directly as a `json blob`. 

**2. Allow for flexibility** 
  - If I really want to load bootstrap over GCPR encoded as thrift, I should be able to do that. 
  - We do not want to limit how you load the data, but we want to control when you load it. 

**3. Make it fast**
   - We need the ability to things in parallel. The response from the Unleash API should always win, but it is enough for the bootstrap to return to consider the SDK as ready to use".

I believe currently the Ruby SDK is blocking until it has received an initial response from the API. This is OK for now, and means we need to load the bootstrap before loading from the Unleash API, and if the bootstrap succeed and the Unleash API is unavailable, the SDK should still return, and operate on the "bootstrap data". 

**We do however want the opportunity in the future to load both the bootstrap and call the Unleash API in parallel on initialization, and then we will need to be in control on how to call.** 

**Example from Node.js SDK ([link](https://github.com/Unleash/unleash-client-node#bootstrap))**
In order to accommodate 1., 2. and  3. in the Node SDK we ended up with allowing the user to provide multiple optional parameters to the SDK:

- `boostrap.url` - A url to fetch the bootstrap data from. 
- `boostrap.urlHeaders` - Headers to add when fetching the `boostrap.url`
- `boostrap.file` - A file to load the bootstrap data from
- `boostrap.data` - A object structure for the bootstrap data (js supports json nativity)
- `bootstrapProvider` - A class implementing a defined interface with a single async function the SDK can call to load the bootstrap. In hindsight I see that we could solved that with a lambda, but that could also limit future needs, if we would like more methods. 

The Node SDK will not expect the user to set all these options simultaneously, and has a strict load preference (data > url > file)  . But requiring the user to just have to worry about _the location of the bootstrap_, and not the loading itself, is an exreeme simplification (1) for the standard use cases. And if the built-in alternatives does not fit the bill they can always provide their own `bootstrapProvider` (2).

PR #88, opened 2022-02-10
Bootstrapping - another approach

sighphyre:


README.md
```diff
@@ -268,6 +273,53 @@ variant = UNLEASH.get_variant "ColorVariants", @unleash_context, fallback_varian
...
+curl -H 'Authorization: <API token>' -XGET 'http://unleash.herokuapp.com/api' > ./default-toggles.json
+```
+
+Now using them on start up:
+
+```ruby
+
+custom_boostrapper = -> {
+  File.read('./default-toggles.json')
+}
```
rarruda, line 310: I would like to suggest that in the examples we show the `Proc` syntax:

```suggestion
custom_boostrapper = Proc.new {
  File.read('./default-toggles.json')
}
```

sighphyre: Yeah sure! That looks more obvious

examples/bootstrap.rb
```diff
@@ -0,0 +1,53 @@
...
+  app_name: 'bootstrap-test',
+  instance_id: 'local-test-cli',
+  refresh_interval: 2,
+  disable_client: true,
+  disable_metrics: true,
+  metrics_interval: 2,
+  retry_limit: 2,
+  bootstrap_config: Unleash::Bootstrap::Configuration.new({
+                                                            file_path: "examples/default-toggles.json"
+                                                          })
```
rarruda, line 21:
```suggestion
  bootstrap_config: Unleash::Bootstrap::Configuration.new(file_path: "examples/default-toggles.json")
```

README.md
```diff
@@ -268,6 +273,53 @@ variant = UNLEASH.get_variant "ColorVariants", @unleash_context, fallback_varian
...
+    bootstrap_config: Unleash::Bootstrap::Configuration.new({
+        url: "http://unleash.herokuapp.com/api/client/features",
+        url_headers: {'Authorization': '<API token>'}
+    })
+)
+```
+The `Bootstrap::Configuration` initializer takes a hash with one of the following options specified:
+
+* `file_path` - An absolute or relative path to a file containing a JSON string of the response body from the Unleash server. This can also be set though the `UNLEASH_BOOTSTRAP_FILE` environment variable.
+* `url` - A url pointing to an Unleash server's features endpoint, the code sample above is illustrative. This can also be set though the `UNLEASH_BOOTSTRAP_URL` environment variable. If using the `url` parameter, a second `url_headers` hash can be passed, which the bootstrapper will pass to the server, see the code sample above for an example. If this option isn't set then the bootstrapper will use the same url headers as the Unleash client.
```
rarruda, line 293:
```suggestion
* `url` - A url pointing to an Unleash server's features endpoint, the code sample above is illustrative. This can also be set though the `UNLEASH_BOOTSTRAP_URL` environment variable. 
* `url_headers` - Headers for the GET http request to the `url` above. Only used if the `url` parameter is also set. If this option isn't set then the bootstrapper will use the same url headers as the Unleash client.
```

lib/unleash/scheduled_executor.rb
```diff
@@ -1,19 +1,22 @@
 module Unleash
   class ScheduledExecutor
-    attr_accessor :name, :interval, :max_exceptions, :retry_count, :thread
+    attr_accessor :name, :interval, :max_exceptions, :retry_count, :thread, :immediate_execution
 
-    def initialize(name, interval, max_exceptions = 5)
+    def initialize(name, interval, max_exceptions = 5, immediate_execution = false)
```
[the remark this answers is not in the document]
sighphyre: @rarruda  Not sure how you feel about this one. Looks clear enough to me as is and implementing the suggestion locally breaks my syntax highlighter for some reason, which always makes me nervous

lib/unleash/toggle_fetcher.rb
```diff
@@ -126,5 +127,23 @@ def read!
         file&.close
       end
     end
+
+    def bootstrap
+      unleash_response = Unleash::Bootstrap::Handler.new(Unleash.configuration.bootstrap_config).retrieve_toggles
+      synchronize_with_local_cache! get_features unleash_response
```
rarruda, line 133: So that the variable is more descriptive:
```suggestion
      bootstrap_payload = Unleash::Bootstrap::Handler.new(Unleash.configuration.bootstrap_config).retrieve_toggles
      synchronize_with_local_cache! get_features bootstrap_payload
```

lib/unleash/scheduled_executor.rb
```diff
@@ -1,19 +1,22 @@
 module Unleash
   class ScheduledExecutor
-    attr_accessor :name, :interval, :max_exceptions, :retry_count, :thread
+    attr_accessor :name, :interval, :max_exceptions, :retry_count, :thread, :immediate_execution
```
rarruda, line 3: TBH, I don't like this approach, but we can let it slide for now, if we can refactor in another PR.

sighphyre: I have mixed feelings, that has a few impacts I'm not sure about - it means the metrics task loses its ability to have a delayed start and we can only ever eagerly resolve toggles. That's quite a few knock on effects. But I think we should take that to another PR and talk what/why

lib/unleash/bootstrap/handler.rb
```diff
@@ -0,0 +1,22 @@
...
+      def initialize(configuration)
+        self.configuration = configuration
+      end
+
+      # @return [String] JSON string representing data returned from an Unleash server
+      def retrieve_toggles
+        return Provider::FromFile.read(configuration.file_path) unless self.configuration.file_path.nil?
+        return Provider::FromUrl.read(configuration.url, configuration.url_headers) unless self.configuration.url.nil?
+        return configuration.data unless self.configuration.data.nil?
+        return configuration.closure.call if self.configuration.closure.is_a?(Proc)
```
rarruda, line 18: We should also document the prio order in the README!


```suggestion
        return configuration.data unless self.configuration.data.nil?
        return configuration.closure.call if self.configuration.closure.is_a?(Proc)
        return Provider::FromFile.read(configuration.file_path) unless self.configuration.file_path.nil?
        return Provider::FromUrl.read(configuration.url, configuration.url_headers) unless self.configuration.url.nil?
```

rarruda:
known missing things/to be followed up:
- test ordering of priorities for the bootstrapping config
- refactoring of the scheduled executor

lib/unleash/bootstrap/configuration.rb
```diff
@@ -0,0 +1,25 @@
...
+      def initialize(opts = {})
+        self.file_path = self.resolve_value_indifferently(opts, 'file_path') || ENV['UNLEASH_BOOTSTRAP_FILE'] || nil
+        self.url = self.resolve_value_indifferently(opts, 'url') || ENV['UNLEASH_BOOTSTRAP_URL'] || nil
+        self.url_headers = self.resolve_value_indifferently(opts, 'url_headers')
+        self.data = self.resolve_value_indifferently(opts, 'data')
+        self.closure = self.resolve_value_indifferently(opts, 'closure')
+      end
+
+      def valid?
+        !(self.data || self.file_path || self.url || self.closure).nil?
```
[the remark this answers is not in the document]
rarruda: Yeah, that's more idiomatic/easier to read.

Another option is to use this other syntax:
```suggestion
        %w[self.data self.file_path self.url self.closure].all?(:nil?)
```

lib/unleash/scheduled_executor.rb
rarruda: I agree that we should take this in a separate PR.

lib/unleash/scheduled_executor.rb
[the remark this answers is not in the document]
rarruda: let's leave it as is for now.

We will discuss `immediate_execution` in a separate PR anyhow.

lib/unleash/toggle_fetcher.rb
```diff
@@ -13,10 +14,15 @@ def initialize
...
 
-      # start by fetching synchronously, and failing back to reading the backup file.
       begin
-        fetch
+        # if bootstrap configuration is available, initialize. An immediate API read is also triggered
+        if Unleash.configuration.use_bootstrap?
+          bootstrap
+        else
+          fetch
+        end
```
rarruda, line 23: We still want to fetch at once, even if we read from the bootstrap config, no?

```suggestion
        # if bootstrap configuration is available, initialize. An immediate API read is also triggered
        bootstrap if Unleash.configuration.use_bootstrap?
        fetch
```

lib/unleash/bootstrap/handler.rb
sighphyre: Good suggestion! I've added a warning to the README that you should only select one, along with a description of how they're resolved if multiple are specified

lib/unleash/toggle_fetcher.rb
rarruda: If we do this, then we should be able to drop the changes to the `ScheduledExecutor`, no? (regarding `immediate_execution`)

It means it will block until we fetch the data from the bootstrap but that sounds like an acceptable trade-off?

README.md
```diff
@@ -80,17 +80,22 @@ Argument | Description | Required? |  Type |  Default Value|
 `backup_file` | Filename to store the last known state from the Unleash server. Best to not change this from the default. | N | String | `Dir.tmpdir + "/unleash-#{app_name}-repo.json` |
 `logger` | Specify a custom `Logger` class to handle logs for the Unleash client. | N | Class | `Logger.new(STDOUT)` |
 `log_level` | Change the log level for the `Logger` class. Constant from `Logger::Severity`. | N | Constant | `Logger::WARN` |
+`bootstrap_config` | Bootstrap config on how to loaded data on start-up. This is useful for loading large states on startup without (or before) hitting the network. | N | Unleash::Bootstrap::Configuration | `nil` |
 
 For in a more in depth look, please see `lib/unleash/configuration.rb`.
```
chriswk, line 85:
```suggestion
For a more in-depth look, please see `lib/unleash/configuration.rb`.
```

lib/unleash/toggle_fetcher.rb
sighphyre: So I'm open to suggestion here but as written this won't work - we'll need to spawn off a thread or other concurrency primitive here to not block the caller's thread. Currently the only thing that really does threading is the ScheduledExecutor. Threading is evil and subtle and I'd rather keep it to one place if possible

lib/unleash/bootstrap/configuration.rb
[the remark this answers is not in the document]
sighphyre: Fair enough, it's such a small dataset I'm not going to worry initialising an array and iterating it

lib/unleash/bootstrap/handler.rb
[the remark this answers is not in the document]
sighphyre: Good point, handled in a comment below

lib/unleash/scheduled_executor.rb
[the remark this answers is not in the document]
sighphyre: Sounds good!

lib/unleash/toggle_fetcher.rb
rarruda: You have a point that we don't want to block on this method if the bootstrap config is provided.

I'm however still convinced we can do this in a better way, but we can leave the improvements for a separate PR.

spec/unleash/scheduled_executor_spec.rb
```diff
@@ -49,4 +49,46 @@
 
     Unleash.configuration.instance_id = original_instance_id
   end
+
+  # These two tests are super flaky because they're checking if threading works
+  # We could extend the times to make them less flaky but that would mean slower tests so I'm disabling them for now
```
rarruda, line 54: Tests that include `sleep` are brittle by definition. The tests should be rewritten, and actually we already agreed to re-evaluate `immediate_execution`.

README.md
```diff
@@ -268,6 +273,62 @@ variant = UNLEASH.get_variant "ColorVariants", @unleash_context, fallback_varian
...
+    })
+)
+```
+The `Bootstrap::Configuration` initializer takes a hash with one of the following options specified:
+
+* `file_path` - An absolute or relative path to a file containing a JSON string of the response body from the Unleash server. This can also be set though the `UNLEASH_BOOTSTRAP_FILE` environment variable.
+* `url` - A url pointing to an Unleash server's features endpoint, the code sample above is illustrative. This can also be set though the `UNLEASH_BOOTSTRAP_URL` environment variable.
+* `url_headers` - Headers for the GET http request to the `url` above. Only used if the `url` parameter is also set. If this option isn't set then the bootstrapper will use the same url headers as the Unleash client.
+* `data` - A raw JSON string as returned by the Unleash server.
+* `closure` - A lambda containing custom logic if you need it, an example is provided below.
```
rarruda, line 296: May I suggest that we call it `block`?
```suggestion
* `block` - A lambda containing custom logic if you need it, an example is provided below.
```

That's a more common way to call closures in ruby.

spec/unleash/scheduled_executor_spec.rb
sighphyre: Agreed, they're disabled for now, they were useful for local dev but I don't want to patch them if we're going to rewrite that section in a week

README.md
sighphyre: Done!

rarruda: We should consider making a table out of this:

```suggestion
Argument | Description | Type | Default Value
-- | -- | -- | -- | --
`file_path` | An absolute or relative path to a file containing a JSON string of the response body from the Unleash server | `String` | ENV['UNLEASH_BOOTSTRAP_FILE'] || nil
`url` | A url pointing to an Unleash server's features endpoint for a GET request. | `String` | ENV['UNLEASH_BOOTSTRAP_FILE'] || nil
`url_headers` | Headers for the GET http request to the `url` above. Only used if the `url` parameter is also set. | `Hash` | Same as the Unleash client
`data`  | A raw JSON string with the features (as returned by the Unleash server). | `string` | nil
`block` | A lambda containing custom logic if you need it, an example is provided below. | `Proc` | nil
```

rarruda (approved):
Can you write an issue with the sticking points in this PR so we don't forget them and get to revisit it in due time?

That I remember right now:
- test ordering of priorities for the bootstrapping config
- refactoring of the scheduled executor
- test all different bootstrapping capabilities from the client perspective (Client class)
- test loading bootstrap configuration from environment variables.

Possibly: (for bonus points)
- allow the cli client to use the bootstrapper
- simpler language in the README for bootstrap priorities

PR #92, opened 2022-03-01
Feature/advanced constraints

sighphyre:
This implements extended support for constraint operators, reference implementation on the node SDK is here: https://github.com/Unleash/unleash-client-node/pull/289/files


lib/unleash/constraint.rb
```diff
@@ -1,26 +1,62 @@
...
+    VALID_OPERATORS = [
+      ConstraintMatcher::ContainsConstraint::OPERATORS,
+      ConstraintMatcher::StringConstraint::OPERATORS,
+      ConstraintMatcher::NumericConstraint::OPERATORS,
+      ConstraintMatcher::DateConstraint::OPERATORS,
+      ConstraintMatcher::SemverConstraint::OPERATORS
+    ].flatten.freeze
 
-    def initialize(context_name, operator, values = [])
+    def initialize(context_name, operator, value = [], inverted = false, case_insensitive = false)
```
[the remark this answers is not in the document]
sighphyre: Done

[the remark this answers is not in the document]
sighphyre: Done

spec/unleash/constraint_spec.rb
```diff
@@ -72,5 +72,306 @@
...
+        session_id: 'verylongsesssionid',
+        remote_address: '127.0.0.1',
+        properties: {
+          env: '3.141'
+        }
+      }
+      context = Unleash::Context.new(context_params)
+
+      constraint = Unleash::Constraint.new('env', 'NUM_LT', '2.718')
+      expect(constraint.matches_context?(context)).to be_falsey
```
rarruda, line 168: I think it is fair that we expect these new tests to return strict `true` or `false` values, no?
```suggestion
      expect(constraint.matches_context?(context)).to be false
```

spec/unleash/constraint_spec.rb
```diff
@@ -72,5 +72,306 @@
       constraint = Unleash::Constraint.new('user_id', 'NOT_IN', ['456', '789'])
       expect(constraint.matches_context?(context)).to be_truthy
     end
+
+    it 'matches based on property STR_STARTS_WITH value' do
+      context_params = {
+        user_id: '123',
+        session_id: 'verylongsesssionid',
+        remote_address: '127.0.0.1',
```
rarruda, line 80: Remove all fields that are not actively tested.
```suggestion
```

lib/unleash/context.rb
```diff
@@ -12,6 +12,7 @@ def initialize(params = {})
       self.user_id     = value_for('userId', params)
       self.session_id  = value_for('sessionId', params)
       self.remote_address = value_for('remoteAddress', params)
+      self.current_time = value_for('currentTime', params, Time.now.utc.iso8601.to_s)
```
rarruda, line 15: Do we know if we want to use current_time as the time that the context object is created or when the constraint is evaluated?

spec/unleash/constraint_spec.rb
```diff
@@ -72,5 +72,306 @@
...
+        properties: {
+          env: '3.141'
+        }
+      }
+      context = Unleash::Context.new(context_params)
+
+      constraint = Unleash::Constraint.new('env', 'NUM_GT', '2.718')
```
rarruda, line 209: maybe it could be nice to have extra scenarios when using higher precision (say 4, or 6 decimal cases), so we also confirm how float rounding is done as expected?

spec/unleash/constraint_spec.rb
```diff
@@ -72,5 +72,306 @@
...
+      context_params = {
+        user_id: '123',
+        session_id: 'verylongsesssionid',
+        remote_address: '127.0.0.1',
+        currentTime: '2022-01-30T13:00:00.000Z'
+      }
+      context = Unleash::Context.new(context_params)
+
+      constraint = Unleash::Constraint.new('currentTime', 'DATE_AFTER', '2022-01-29T13:00:00.000Z')
+      expect(constraint.matches_context?(context)).to be_truthy
```
rarruda, line 295: Would be nice to add more cases?

```suggestion
      constraint = Unleash::Constraint.new('currentTime', 'DATE_AFTER', '2022-01-29T13:00:00.000Z')
      expect(constraint.matches_context?(context)).to be true

      constraint = Unleash::Constraint.new('currentTime', 'DATE_AFTER', '2022-01-29T13:00:00Z')
      expect(constraint.matches_context?(context)).to be true

      constraint = Unleash::Constraint.new('currentTime', 'DATE_AFTER', '2022-01-29T13:00Z')
      expect(constraint.matches_context?(context)).to be true
      
      constraint = Unleash::Constraint.new('currentTime', 'DATE_AFTER', '2022-01-30T12:59:59.999999Z')
      expect(constraint.matches_context?(context)).to be true

      constraint = Unleash::Constraint.new('currentTime', 'DATE_AFTER', '2022-01-30T12:59:59.999Z')
      expect(constraint.matches_context?(context)).to be true
      
      constraint = Unleash::Constraint.new('currentTime', 'DATE_AFTER', '2022-01-30T12:59:59')
      expect(constraint.matches_context?(context)).to be true

      constraint = Unleash::Constraint.new('currentTime', 'DATE_AFTER', '2022-01-30T12:59')
      expect(constraint.matches_context?(context)).to be true
      
      constraint = Unleash::Constraint.new('currentTime', 'DATE_AFTER', '2022-01-30T13:00:00.000Z')
      expect(constraint.matches_context?(context)).to be false
```

spec/unleash/constraint_spec.rb
```diff
@@ -72,5 +72,306 @@
...
+      }
+      context = Unleash::Context.new(context_params)
+      constraint = Unleash::Constraint.new('env', 'STR_STARTS_WITH', ['DEV'], case_insensitive: true)
+      expect(constraint.matches_context?(context)).to be_truthy
+
+      constraint = Unleash::Constraint.new('env', 'STR_ENDS_WITH', ['MENT'], case_insensitive: true)
+      expect(constraint.matches_context?(context)).to be_truthy
+
+      constraint = Unleash::Constraint.new('env', 'STR_CONTAINS', ['LOP'], case_insensitive: true)
+      expect(constraint.matches_context?(context)).to be_truthy
```
rarruda, line 334: Also test when the value is a string, and not just a list of strings.
```suggestion
      constraint = Unleash::Constraint.new('env', 'STR_CONTAINS', ['LOP'], case_insensitive: true)
      expect(constraint.matches_context?(context)).to be_truthy
      
      constraint = Unleash::Constraint.new('env', 'STR_CONTAINS', 'LOP', case_insensitive: true)
      expect(constraint.matches_context?(context)).to be_truthy
```

lib/unleash/constraint.rb
```diff
@@ -1,26 +1,62 @@
...
+    VALID_OPERATORS = [
+      ConstraintMatcher::ContainsConstraint::OPERATORS,
+      ConstraintMatcher::StringConstraint::OPERATORS,
+      ConstraintMatcher::NumericConstraint::OPERATORS,
+      ConstraintMatcher::DateConstraint::OPERATORS,
+      ConstraintMatcher::SemverConstraint::OPERATORS
+    ].flatten.freeze
 
-    def initialize(context_name, operator, values = [])
+    def initialize(context_name, operator, value = [], inverted: false, case_insensitive: false)
```
rarruda, line 20: Since the constructor changed, it's initialization with all parameters should also be tested in the `feature_toggle_spec.rb`

rarruda: Also, methods/constructors with more than 3 parameters are usually frowned upon. It's late so I wont' come with a suggestion right now 🙈 .

rarruda (commented):
I have created a branch (with a single commit) called feature/advanced-constraints-refactored with some (but NOT all) of the suggestions here implemented. 

Mostly because it felt like it was simpler to explain how the end result should look like by creating the branch, and it should be taken as a discussion point.

I however think that some of the suggestions would make for much simpler code. WDYT?

lib/unleash/constraints/contains_constraints.rb
```diff
@@ -0,0 +1,19 @@
...
+      ].freeze
+
+      def self.matches?(operator, context_value, constraint_value)
+        is_included = constraint_value.include? context_value
+        operator == 'IN' ? is_included : !is_included
+      end
+
+      def self.include?(operator)
+        OPERATORS.include? operator
+      end
```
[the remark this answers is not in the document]
sighphyre: It just made the code in constraints not bleed off the edge of my screen but fair enough, if you think it's weird I'll take it out - I think it's a moot point with your refactor, there's some cool stuff in there

spec/unleash/constraint_spec.rb
sighphyre: So this is my bad, I don't think the code is reflecting the intent properly here. This operator shouldn't ever receive single values and an operation on them is undefined in the spec. I think I'm going to take your suggestion using the dispatch in your branch and see if I can make this case impossible to get to with some proper validation

sighphyre:
@rarruda I see your branch and your link. Very cool! I'm going to run with this and see if I can use it to clean this up. I want to add in the warn on failing operators and put in validation that makes it more obvious what operators can/should be used with what values

lib/unleash/constraint.rb
sighphyre: Yes! I'll extend the tests here! I'm open to suggestions on the method parameters but a value object felt like overkill here. Hash maybe?

lib/unleash/context.rb
sighphyre: Such a good question! I see that the SDKs do different things here. I was working off the Java implementation which does this but I see the NodeSDK evaluates the current time during constraint evaluation. I think you're right here (if I'm reading between the lines correctly!) - doing it during evaluation is more correct

lib/unleash/constraint.rb
```diff
@@ -1,26 +1,62 @@
...
+      elsif ConstraintMatcher::NumericConstraint.include? self.operator
+        ConstraintMatcher::NumericConstraint.matches?(self.operator, context_value, self.value)
+      elsif ConstraintMatcher::DateConstraint.include? self.operator
+        ConstraintMatcher::DateConstraint.matches?(self.operator, context_value, self.value)
+      elsif ConstraintMatcher::SemverConstraint.include? self.operator
+        ConstraintMatcher::SemverConstraint.matches?(self.operator, context_value, self.value)
+      else
+        Unleash.logger.warn "Invalid constraint operator: #{self.operator}, this should be unreachable. Defaulting to false"
+        false
+      end
```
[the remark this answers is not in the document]
sighphyre: Your suggestion in that branch is awesome. I actually want to put in a second method here on each class for more aggressive validation. Let me a try a thing and see if you like it, if you still think it's wonky we can move closer to your suggestion

lib/unleash/constraint.rb
```diff
@@ -1,26 +1,62 @@
...
       raise ArgumentError, "operator does not hold a valid value:" + VALID_OPERATORS unless VALID_OPERATORS.include? operator
-      raise ArgumentError, "values does not hold an Array" unless values.is_a?(Array)
+      raise ArgumentError, "value must either hold an array or a single string" unless value.is_a?(Array) || value.is_a?(String)
 
       self.context_name = context_name
       self.operator = operator
-      self.values = values
+      self.value = value
+      self.inverted = inverted
+      self.case_insensitive = case_insensitive
```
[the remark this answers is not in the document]
rarruda: Also, jsyk (and please don't steal this from me) I want to implement [RBS](https://github.com/ruby/rbs) (also see https://www.honeybadger.io/blog/ruby-rbs-type-annotation/) to make things a bit more explicit.

And from there generate the documentation with signatures for the public/client methods. :)

lib/unleash/constraint.rb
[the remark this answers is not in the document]
rarruda: I don't think you need a second method in the classes. you can get around with a bit of passing blocks around:

I think with a few methods you can do that, something like: (warning: not tested)
```ruby
def on_valid_float(val1, val2, &blk)
  begin
    val1 = Float(val1)
    val2 = Float(val2)
  rescue ArgumentError
    Unleash.logger.warn "Unleash::ConstraintMatcher unable to coalesce either context_value (#{val1}) or constraint_value (#{val2}) into a number. Returning always false."
    return false
  end

  yield (val1, val2)
end
```

then you can do:
```ruby
  STR_EQ: lambda {|v1,v2| validate_str(v1,v2} { 
  NUM_EQ: lambda { |context_value, constraint_value| on_valid_float(constraint_value, context_value) { (Float(constraint_value) - Float(context_value)).abs < 0.001 } },
```

and different methods for each validation that you need.

I can create working examples later tonight.

lib/unleash/constraint.rb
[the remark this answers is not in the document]
sighphyre: I feel so bad now! I promise I'll let you have some features. That looks pretty awesome!

lib/unleash/constraint.rb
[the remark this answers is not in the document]
sighphyre: Thanks, this worked really nicely!

lib/unleash/constraints/contains_constraints.rb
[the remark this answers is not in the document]
sighphyre: This was made irrelevant by other suggestions here, going to close this one off

lib/unleash/constraints/numeric_constraints.rb
```diff
@@ -0,0 +1,39 @@
...
+        'NUM_LT',
+        'NUM_LTE'
+      ].freeze
+
+      def self.matches?(operator, context_value, constraint_value)
+        begin
+          context_value = Float(context_value)
+          constraint_value = Float(constraint_value)
+        rescue ArgumentError
+          false
```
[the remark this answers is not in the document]
sighphyre: Done!

lib/unleash/constraints/contains_constraints.rb
```diff
@@ -0,0 +1,19 @@
...
+    class ContainsConstraint
+      OPERATORS = [
+        'IN',
+        'NOT_IN'
+      ].freeze
+
+      def self.matches?(operator, context_value, constraint_value)
+        is_included = constraint_value.include? context_value
+        operator == 'IN' ? is_included : !is_included
+      end
```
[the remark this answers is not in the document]
sighphyre: Thanks, done!

spec/unleash/constraint_spec.rb
sighphyre: So I've added a test case checking if (0.1 + 0,2).to_s is numerically equal to 0.3, this should be one of the edge cases where the computation doesn't match 0,3 exactly but should still resolve correctly. I don't want to add too many more because it feels like we're checking Ruby's floating point spec and we're checking against Float::EPSILON in the lambda. But I'm open to adding if folks feel strongly about it

spec/unleash/constraint_spec.rb
sighphyre: So the case described here is actually undefined in the spec and the code was too permissive. I've added validation in the constraint constructor to reject invalid datatypes

spec/unleash/constraint_spec.rb
sighphyre: Good call, done!

lib/unleash/context.rb
sighphyre: So given that the other SDKs do this, I want to leave this alone. Thoughts?

lib/unleash/constraint.rb
sighphyre: I've been thinking about this. I don't think a Hash is better here, it makes the values opaque to the caller and requires extra validation layers and two of the parameters are optionals, for most cases the calling code shouldn't be excessively long. I'm tempted to just leave this as is but if anyone has a better idea than Hashes (or can tell me I'm wrong about their usage here!) I'm open

lib/unleash/constraint.rb
```diff
@@ -41,7 +41,12 @@ def matches_context?(context)
       Unleash.logger.debug "Unleash::Constraint matches_context? value: #{self.value} context.get_by_name(#{self.context_name})" \
         " #{context.get_by_name(self.context_name)} "
       match = matches_constraint?(context)
+      puts "CHECKING A CONSTRAIT #{match}, #{self.operator}, #{context}"
       self.inverted ? !match : match
+    rescue KeyError
+      Unleash.logger.debug "Attemped to resolve a context key during constraint resolution: #{self.context_name} but it wasn't \
```
rarruda, line 46: i would suggest making this a WARN, instead of DEBUG

spec/unleash/constraint_spec.rb
```diff
@@ -61,16 +61,358 @@
...
+
+      constraint = Unleash::Constraint.new('env', 'NUM_LT', '6.282')
+      expect(constraint.matches_context?(context)).to be true
+    end
+
+    it 'matches based on property NUM_LTE value' do
+      context_params = {
+        user_id: '123',
+        session_id: 'verylongsesssionid',
+        remote_address: '127.0.0.1',
```
rarruda, line 181:
```suggestion
```

spec/unleash/constraint_spec.rb
```diff
@@ -61,16 +61,358 @@
...
+
+      constraint = Unleash::Constraint.new('env', 'NUM_EQ', (0.1 + 0.2).to_s)
+      expect(constraint.matches_context?(context)).to be true
+    end
+
+    it 'matches based on property NUM_LT value' do
+      context_params = {
+        user_id: '123',
+        session_id: 'verylongsesssionid',
+        remote_address: '127.0.0.1',
```
rarruda, line 160:
```suggestion
```

spec/unleash/constraint_spec.rb
```diff
@@ -61,16 +61,358 @@
...
+
+      constraint = Unleash::Constraint.new('env', 'STR_CONTAINS', ['DEVELOPMENT'])
+      expect(constraint.matches_context?(context)).to be false
+    end
+
+    it 'matches based on property NUM_EQ value' do
+      context_params = {
+        user_id: '123',
+        session_id: 'verylongsesssionid',
+        remote_address: '127.0.0.1',
```
rarruda, line 140:
```suggestion
```

spec/unleash/constraint_spec.rb
```diff
@@ -61,16 +61,358 @@
...
+        properties: {
+          env: '0.3'
```
rarruda, line 142: Bikeshedding: 

Maybe use a different var name than `env` ?
```suggestion
          distance: '0.3'
```

spec/unleash/constraint_spec.rb
```diff
@@ -61,16 +61,358 @@
...
+
+      constraint = Unleash::Constraint.new('env', 'STR_ENDS_WITH', ['dev'])
+      expect(constraint.matches_context?(context)).to be false
+    end
+
+    it 'matches based on property STR_CONTAINS value' do
+      context_params = {
+        user_id: '123',
+        session_id: 'verylongsesssionid',
+        remote_address: '127.0.0.1',
```
rarruda, line 117:
```suggestion
```

spec/unleash/constraint_spec.rb
```diff
@@ -61,16 +61,358 @@
...
+
+      constraint = Unleash::Constraint.new('env', 'STR_STARTS_WITH', ['ment'])
+      expect(constraint.matches_context?(context)).to be false
+    end
+
+    it 'matches based on property STR_ENDS_WITH value' do
+      context_params = {
+        user_id: '123',
+        session_id: 'verylongsesssionid',
+        remote_address: '127.0.0.1',
```
rarruda, line 97:
```suggestion
```

lib/unleash/constraint.rb
rarruda: I feel quite uneasy about the hybrid approach here. (using hashes for the new values, but not for everything. But again, maybe we can just leave as is for the params w/o default values?

rarruda (commented):
This looks good, but I left a few extra comments on where things could be even better.
But note that many comments are bike shedding, and not that important.

Also, please try to fix the rubocop warnings before we merge to master.

For example: bump rubocop to allow classes wih 122 lines :)

But all in all, this is looking great!

lib/unleash/constraint.rb
sighphyre: Good spot! Done

lib/unleash/constraint.rb
```diff
@@ -1,26 +1,104 @@
+require 'date'
+
 module Unleash
   class Constraint
-    attr_accessor :context_name, :operator, :values
+    attr_accessor :context_name, :operator, :value, :inverted, :case_insensitive
+
+    OPERATORS = {
+      IN: ->(context_value, constraint_value){ constraint_value.include? context_value },
```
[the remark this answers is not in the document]
sighphyre: I like it, not bike shedding at all! This code is doing its best to bleed off the edge of my screen, the small rename makes it all visible without scrolling - Done!

lib/unleash/constraint.rb
```diff
@@ -1,26 +1,104 @@
...
+      NUM_GTE: ->(context_value, constraint_value){ on_valid_float(constraint_value, context_value){ |x, y| (x <= y) } },
+      DATE_AFTER: ->(context_value, constraint_value){ on_valid_date(constraint_value, context_value){ |x, y| (x < y) } },
+      DATE_BEFORE: ->(context_value, constraint_value){ on_valid_date(constraint_value, context_value){ |x, y| (x > y) } },
+      SEMVER_EQ: ->(context_value, constraint_value){ on_valid_version(constraint_value, context_value){ |x, y| (x == y) } },
+      SEMVER_GT: ->(context_value, constraint_value){ on_valid_version(constraint_value, context_value){ |x, y| (x < y) } },
+      SEMVER_LT: ->(context_value, constraint_value){ on_valid_version(constraint_value, context_value){ |x, y| (x > y) } }
+    }.freeze
 
-    VALID_OPERATORS = ['IN', 'NOT_IN'].freeze
+    VALID_LIST_TYPES = ["IN", "NOT_IN", "STR_STARTS_WITH", "STR_ENDS_WITH", "STR_CONTAINS"].map(&:to_sym).freeze
```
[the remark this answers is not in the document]
sighphyre: Yeah LIST_OPERATORS is better, definitely reads nicer now - also shifted this to syms

lib/unleash/constraint.rb
sighphyre: I have thought about this one way too much - yeah the hash suggestion isn't right here. I want to leave the new default values in for the inverted and case Insensitive, mostly because they won't be needed 95% of the time (caseInsensitive won't do anything outside of the string operators and most constraints won't need inversion because they can be expressed as their equal and opposite constraint - !GTE == LT)

PR #93, opened 2022-03-23
correct logic for default values on feature toggles

shieldo:
The incumbent behaviour of feature toggles when they are present but disabled, and the default value was set to true, seemed to be incorrect - it would resolve to true, using the default value, rather than the value of the existing toggle. This meant that, for example (in the case of using the default strategy), code like `UNLEASH.is_enabled?('my-feature-toggle', context, true)` will always resolve to true regardless of the value of the toggle itself.

It doesn't seem correct that code within the feature toggle resolution code was taking default values - if a feature toggle is present, the default value shouldn't be relevant, so far as I understand the intent behind specifying a default value.

This PR corrects the assertions made in the tests for feature toggles, and corrects the behaviour accordingly. Please feed back If I've misunderstood anything here!

I would note also that given systems may be relying on this behaviour, even accidentally, in their applications, acceptance of this PR is a potential major backwards compatibility break, and therefore I would suggest a new major version release.

### DIFF
--- a/lib/unleash/client.rb
+++ b/lib/unleash/client.rb
@@ -47,7 +47,7 @@ module Unleash
 
       toggle = Unleash::FeatureToggle.new(toggle_as_hash)
 
-      toggle.is_enabled?(context, default_value)
+      toggle.is_enabled?(context)
     end
 
     # enabled? is a more ruby idiomatic method name than is_enabled?
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -23,8 +23,8 @@ module Unleash
       "<FeatureToggle: name=#{name},enabled=#{enabled},strategies=#{strategies},variant_definitions=#{variant_definitions}>"
     end
 
-    def is_enabled?(context, default_result)
-      result = am_enabled?(context, default_result)
+    def is_enabled?(context)
+      result = am_enabled?(context)
 
       choice = result ? :yes : :no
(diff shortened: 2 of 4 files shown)


sighphyre (changes requested):
Ah! Thanks for this and nice catch. Philosophically I agree with the change expressed here, this does seem to be a bug, based on what's in the docs: https://docs.getunleash.io/client-specification#:~:text=Calling%20unleash%20with%20default%20value,boolean%20value%20%3D%20unleash. Also hugely appreciate the updates to tests!

This change has, however, broken the specification tests, I'm pretty sure this is because of internal API changes and not as a result of invalid logic but we'll need to get the specifications running before we can look at merging this. They can be pulled into the project with the following shell command:

`git clone --depth 5 --branch v4.1.0 https://github.com/Unleash/client-specification.git client-specification`

Then `rake spec` will pick them up and run them. Once we have those green we can make sure this hasn't changed anything that it wasn't supposed to

shieldo:
@sighphyre Thanks for the feedback! I've updated the client specification test so that it runs green - as you suggested, they were only failures due to the internal API change.

sighphyre:
@shieldo Thanks so much for doing that! This looks good to me, I'm going to leave this open for the day so that others can take a glance but if there's no comments by close of day I'll merge it in.

sighphyre (approved):

rarruda (approved):

rarruda:
Nice catch! Can't really say what I had in mind way back then, and how this persisted for so long!

PR #95, opened 2022-04-15
Quote 3.0 to avoid truncation to 3

petergoldstein:
An unquoted 3.0 is truncated to 3, causing the latest Ruby 3 to be loaded.  At this point that's a 3.1.x version, which is not what's intended.

By quoting the 3.0 we ensure a 3.0.x version is loading for this entry in the CI matrix.  You can see this by looking at a previously run action on this repo:

Screen Shot 2022-04-14 at 6 03 14 PM


rarruda (approved):
Awesome! Thanks for catching this!

PR #96, opened 2022-04-18
Allow custom strategies

volodymyr-mykhailyk:
Other clients and proxy allows implementation and usage of [custom activation strategies](https://docs.getunleash.io/advanced/custom_activation_strategy). This pull request provides easy way for ruby client to be extend with those strategies. 

Previous state:
- Strategies constructed at runtime into a constant.
- Strategies have duplicated logic for strategy name (method, initialization code)
- Strategies matched by converting everything into symbol
- Strategies stored as hash and force clients to work with hash

Pull Request:
- Wraps strategies management mechanism into simple interface
- Using strategy `name` method as single source of identification
- Allows adding new strategies at runtime
- Changes strategy construction from reflection to simple list (default strategies are not changed so often to require magick)

### DIFF
--- a/lib/unleash/strategies.rb
+++ b/lib/unleash/strategies.rb
@@ -1,5 +1,5 @@
 require 'unleash/strategy/base'
-Gem.find_files('unleash/strategy/**/*.rb').each{ |path| require path unless path.end_with? '_spec.rb' }
+Gem.find_files('unleash/strategy/**/*.rb').each{ |path| require path }
 
 module Unleash
   class Strategies
--- a/lib/unleash/strategies.rb
+++ b/lib/unleash/strategies.rb
@@ -7,10 +7,9 @@ module Unleash
       @strategies = Unleash::Strategy.constants
         .select{ |c| Unleash::Strategy.const_get(c).is_a? Class }
         .reject{ |c| ['NotImplemented', 'Base'].include?(c.to_s) }
-        .map do |c|
-        lowered_c = c.to_s
-        lowered_c[0] = lowered_c[0].downcase
-        [lowered_c.to_sym, Object.const_get("Unleash::Strategy::#{c}").new]
+        .map do |strategy_class|
+        strategy = Object.const_get("Unleash::Strategy::#{strategy_class}").new
+        [strategy.name, strategy]
       end.to_h
     end
 
@@ -19,11 +18,11 @@ module Unleash
     end
 
     def includes?(name)
-      @strategies.has_key?(name)
+      @strategies.has_key?(name.to_s)
     end
 
     def fetch(name)
-      raise Unleash::Strategy::NotImplemented, "Strategy is not implemented" unless (strategy = @strategies[name])
+      raise Unleash::Strategy::NotImplemented, "Strategy is not implemented" unless (strategy = @strategies[name.to_s])
 
       strategy
     end
--- a/lib/unleash/strategies.rb
+++ b/lib/unleash/strategies.rb
@@ -5,7 +5,7 @@ module Unleash
   class Strategies
     def initialize
       @strategies = {}
-      DEFAULT_STRATEGIES.each{ |strategy_class| add(strategy_class.new) }
+      register_strategies
     end
 
     def keys
@@ -32,6 +32,26 @@ module Unleash
 
     def [](key)
       @strategies[key.to_s]
+    end
+
+    def register_strategies
+      register_base_strategies
+      register_custom_strategies
+    end
+
+    protected
+
+    def register_custom_strategies
+      Unleash::Strategy.constants
+        .select{ |c| Unleash::Strategy.const_get(c).is_a? Class }
+        .reject{ |c| ['NotImplemented', 'Base'].include?(c.to_s) } # Reject abstract classes
+        .map{ |c| Object.const_get("Unleash::Strategy::#{c}") }
+        .reject{ |c| DEFAULT_STRATEGIES.include?(c) } # Reject base classes
+        .each{ |c| self.add(c.new) }
+    end
+
+    def register_base_strategies
+      DEFAULT_STRATEGIES.each{ |c| self.add(c.new) }
     end
 
     DEFAULT_STRATEGIES = [
--- a/lib/unleash/strategies.rb
+++ b/lib/unleash/strategies.rb
@@ -1,3 +1,6 @@
+require 'unleash/strategy/base'
+Gem.find_files('unleash/strategy/**/*.rb').each{ |path| require path unless path.end_with? '_spec.rb' }
+
 module Unleash
   class Strategies
     def initialize
--- a/lib/unleash/strategies.rb
+++ b/lib/unleash/strategies.rb
@@ -27,6 +27,7 @@ module Unleash
     end
 
     def []=(key, strategy)
+      warn_deprecated_registration(strategy, 'modifying Unleash::STRATEGIES')
       @strategies[key.to_s] = strategy
     end
 
@@ -41,13 +42,18 @@ module Unleash
 
     protected
 
+    # Deprecated: Use Unleash.configuration to add custom strategies
     def register_custom_strategies
       Unleash::Strategy.constants
         .select{ |c| Unleash::Strategy.const_get(c).is_a? Class }
         .reject{ |c| ['NotImplemented', 'Base'].include?(c.to_s) } # Reject abstract classes
         .map{ |c| Object.const_get("Unleash::Strategy::#{c}") }
         .reject{ |c| DEFAULT_STRATEGIES.include?(c) } # Reject base classes
-        .each{ |c| self.add(c.new) }
+        .each do |c|
+        strategy = c.new
+        warn_deprecated_registration(strategy, 'adding custom class into Unleash::Strategy namespace')
+        self.add(strategy)
+      end
     end
 
     def register_base_strategies
@@ -64,5 +70,11 @@ module Unleash
       Unleash::Strategy::RemoteAddress,
       Unleash::Strategy::UserWithId
     ].freeze
+
+    def warn_deprecated_registration(strategy, method)
+      warn "[DEPRECATED] Registering custom Unleash strategy by #{method} is deprecated.
+             Please use Unleash configuration to register custom strategy: " \
+           "`Unleash.configure {|c| c.strategies.add(#{strategy.class.name}.new) }`"
+    end
   end
 end
--- a/lib/unleash/strategies.rb
+++ b/lib/unleash/strategies.rb
@@ -4,13 +4,8 @@ Gem.find_files('unleash/strategy/**/*.rb').each{ |path| require path }
 module Unleash
   class Strategies
     def initialize
-      @strategies = Unleash::Strategy.constants
-        .select{ |c| Unleash::Strategy.const_get(c).is_a? Class }
-        .reject{ |c| ['NotImplemented', 'Base'].include?(c.to_s) }
-        .map do |strategy_class|
-        strategy = Object.const_get("Unleash::Strategy::#{strategy_class}").new
-        [strategy.name, strategy]
-      end.to_h
+      @strategies = {}
+      DEFAULT_STRATEGIES.each{ |strategy_class| add(strategy_class.new) }
     end
 
     def keys
@@ -30,5 +25,16 @@ module Unleash
     def add(strategy)
       @strategies[strategy.name] = strategy
     end
+
+    DEFAULT_STRATEGIES = [
+      Unleash::Strategy::ApplicationHostname,
+      Unleash::Strategy::Default,
+      Unleash::Strategy::FlexibleRollout,
+      Unleash::Strategy::GradualRolloutRandom,
+      Unleash::Strategy::GradualRolloutSessionId,
+      Unleash::Strategy::GradualRolloutUserId,
+      Unleash::Strategy::RemoteAddress,
+      Unleash::Strategy::UserWithId
+    ].freeze
   end
 end
--- a/lib/unleash/strategies.rb
+++ b/lib/unleash/strategies.rb
@@ -26,6 +26,14 @@ module Unleash
       @strategies[strategy.name] = strategy
     end
 
+    def []=(key, strategy)
+      @strategies[key.to_s] = strategy
+    end
+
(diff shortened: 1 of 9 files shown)


rarruda:
I unfortunately won't have the time to look too deep into your PR in the next couple of days. But I have an old branch https://github.com/Unleash/unleash-client-ruby/tree/feature/allow_strategies_from_argument which was a draft attempt to get the same functionality in place. (didn't feel good enough as was to create a PR out of it).

As soon as i have a bit of slack in my other fronts I will get back to you.

rarruda:
Note that also it is possible right now  to use custom strategies either through a [custom gem](https://github.com/rarruda/unleash-client-ruby-strategy-companyid) or monkey patching it in the right namespace. Neither are super good practices.

So having a way to specify it in the constructor of the client, is definitely wanted/on the roadmap.

volodymyr-mykhailyk:
@rarruda, no worries, this is not critical. I know we can do that via monkey patching, and we actually do it on one of the products, but since we want to adopt unleash on more - decided improvement to the gem will be better than copying patches around infrastructure. 

I reviewed your previous version - and instead of working around constant - decided it make more sense to make a list of strategies configurable. Strategies manager class is an attempt to reduce code/complexity when other parts use those strategies and abstract storing mechanisms.

Will be waiting for your review, and do not hesitate to let me know how else I can help.

sighphyre:
@volodymyr-mykhailyk Ah this is awesome! I think we've been wanting to implement this for a while and there's always something else to get to first. I'd like to be quite careful with this one because it's hard to change implementations when they're in use so I need a bit of time to digest and maybe draw some inspiration from @rarruda's spike. I think it's also important to consider backwards compatibility here. I'll definitely raise this with the team on Monday to put this into our planning and get some feedback to you next week. Thanks so much for the PR!

rarruda:
First: 

The core of this PR is quite sound, and does improve well on how to handle this feature. I specially like that we keep the custom strategies in the configuration object.

But some feedback regarding the implementation 
in finer details:

> Wraps strategies management mechanism into simple interface

I feel a simple list of parameters is a more desirable behavior, and more closely maps how the other SDKs behave. I would prefer to pass classes than to add an interface, as no other configuration parameter currently has an interface. Also the strategies class is almost a wrapper around a HashMap, and given that we mostly use that internally in the SDK, and that is not exposed to users, this adds unnecessary complexity.

> Using strategy name method as single source of identification
> Allows adding new strategies at runtime

I don't see this as an important feature. The most important is to be able to able to provide custom strategies when the client is created. It could even be quite bug inducing to allow changing the configuration parameters later on.

> Changes strategy construction from reflection to simple list (default strategies are not changed so often to require magick)

This will likely break other people's code, as this way to add custom strategies is known to be in use already. 


I would recommend trying to reduce the footprint of the PR to a minimum required to add the feature, and in a way that does not break current users way to handle custom strategies.

volodymyr-mykhailyk:
@rarruda Thanks for the review. Let me reply to the few comments and based on your next suggestion I will update the PR.

> > Wraps strategies management mechanism into simple interface
> 
> I feel a simple list of parameters is a more desirable behavior, and more closely maps how the other SDKs behave. I would prefer to pass classes than to add an interface, as no other configuration parameter currently has an interface. Also the strategies class is almost a wrapper around a HashMap, and given that we mostly use that internally in the SDK, and that is not exposed to users, this adds unnecessary complexity.

You questioned `StrategiesManager` existence here, is that right? While I was implementing the code - I was debating myself as well if I should simply expose the list/hash. However, decided that encapsulation will be more appropriate and descriptive in this case. Also, it abstracts details of implementation and serves as a place for code to provide backward compatibility (more on that later):
```
config.strategies.add(MyCustomStrategy.new)
# vs
config.strategies['key'] = MyCustomStrategy.new
```

**Question**: If you think it is still too much abstraction - let me know. Will happily remove wrapper.

> 
> > Using strategy name method as single source of identification
> > Allows adding new strategies at runtime
> 
> I don't see this as an important feature. The most important is to be able to able to provide custom strategies when the client is created. It could even be quite bug inducing to allow changing the configuration parameters later on.

You quoted two items here. Please help me to understand what should be changed
- Using the strategy name method as a single source of identification - Previously in the code - there were two places where gem defined the name for the strategy. [During initialization](https://github.com/Unleash/unleash-client-ruby/blob/main/lib/unleash.rb#L19) by converting the class name into the string. [Second](https://github.com/Unleash/unleash-client-ruby/blob/main/lib/unleash/strategy/base.rb#L7) - name method on strategy class itself, that wasn't used. With this change, I always refer to the strategy name method and removed the custom code that guessed the strategy name from class name. 

**Question**: Do you want me to revert this change? Should we remove the name method on strategy?

- Allows adding new strategies at runtime - Sorry, I was too vague in my description here. I do not intend to provide a way to change the list of strategies after the gem has been initialized/configured. I as well agree this is not very safe to change strategies in the middle of a 'request'. Under this comment, I meant that strategies are now defined not at the [file reading phase](https://github.com/Unleash/unleash-client-ruby/blob/main/lib/unleash.rb#L13), but at runtime when you initialize Unleash. 

**Question**: Do you want to change something about this? Do you want to add protective measures to avoid changing list of strategies after Unleash was initialized?

> > Changes strategy construction from reflection to simple list (default strategies are not changed so often to require magick)
> 
> This will likely break other people's code, as this way to add custom strategies is known to be in use already.
> 
> I would recommend trying to reduce the footprint of the PR to a minimum required to add the feature, and in a way that does not break current users way to handle custom strategies.

Ok. Now, after reviewing your comments and original code again I think I understand what you mean by backward compatibility. Let me first explain why I think the previous way of defining strategies is not desirable in ruby/rails projects, and then will be open to updating the PR in a backward-compatible way vs suggesting releasing a new major version and forcing gem users to migrate.

In the original code - the list of strategies is defined at the moment when ruby is [reading the file](https://github.com/Unleash/unleash-client-ruby/blob/main/lib/unleash.rb#L13). Ruby is creating constant `STRATEGIES` and its content the moment you do `require 'unleash'`. In the ruby project, it is very common to automatically require all your dependencies at the loading phase of your framework (for Rails in order to not load a gem you need to explicitly mark it in the bundler, and then manually require the gem). So it leaves us with only two options on how one can add a custom strategy:

1. Fiddle with code loading order:
  1.1. Mark gem as `require: false` in bundler
  1.2. Create your strategy that is placed in Unleash namespace `Unleash::Strategy::MyStrategy`
  1.3. Explicitly require your strategy file `require 'unleash/strategy/my_strategy`
  1.4. Explicitly require the gem code `require 'unleash'`
  1.5. Initialize unleash
2. Monkey patch contents of the constant
  2.1. Change the constant: `Unleash::STRATEGIES[:MyStrategy] = MyStrategy.new`
  2.2. Initialize Unleash

I find the first option very intrusive in how the ruby/rails project loads code and dependencies and is very easy to brake as this approach is not going to fail if your custom strategy is loaded after unleash was initialized.  While the second option is much easier and more explicit to tell if your custom strategy is not configured - it is still not much better semantically as we changing the *constant* and require users to understand the internal data structure and operation of the gem.  

As you correctly noted - my change is going to break the functionality for everyone. However, considering the hidden risk in the first approach, and "it's just wrong" feelings about the second approach - would welcome the gem provider to encourage me to update integration and make the overall configuration a bit more straightforward. 

I can change the PR to be backward compatible with both methods of adding custom strategy but feeling a bit weird to keep cultivating bad code around

**Question**: Do you want me to make this PR backward compatible or we better release a new major version that will require users to review their unleash configuration code?

sighphyre:
@volodymyr-mykhailyk Thanks for the detailed response, that makes a lot of sense! To address a few questions here:

> Question: If you think it is still too much abstraction - let me know. Will happily remove wrapper.

My vote here is to use a hash over the class. It feels like more abstraction than we currently need (if we can tackle the backwards compatibility conversation without it).

> Question: Do you want me to revert this change? Should we remove the name method on strategy?
My primary concern here is that it sounded like we were moving towards a feature where the user could dynamically update strategies on the fly. From reading your description, now I realise this has more to do with load order and that makes a lot of sense to me. Given that, I don't think we need anything to block updates to the strategies list. I'm happy here, @rarruda might spot something that I've missed though

In terms of the conversation about backwards compatibility I definitely agree, we've known for a while that they way we loaded custom strategies wasn't what we wanted it to be. That being said I have two worries here, firstly there are users in the wild that are currently using that approach and humans are terribly bad at reading patch notes, even in major releases. Secondly, adding breaking changes is going to lock people out of updates here until they can update to make use of the approach in this PR. I want to say that ideally we should provide backwards compatibly for a period of time, which should give us time to warn folks in our docs and give them breathing room to get this updated smoothly. Definitely agree that we should move towards removing the old approach soon. Is it a horrible amount of work to support both for a period of time?

volodymyr-mykhailyk:
> @volodymyr-mykhailyk Thanks for the detailed response, that makes a lot of sense! To address a few questions here:
> 
> > Question: If you think it is still too much abstraction - let me know. Will happily remove wrapper.
> 
> My vote here is to use a hash over the class. It feels like more abstraction than we currently need (if we can tackle the backwards compatibility conversation without it).
> 
> > Question: Do you want me to revert this change? Should we remove the name method on strategy?
> > My primary concern here is that it sounded like we were moving towards a feature where the user could dynamically update strategies on the fly. From reading your description, now I realise this has more to do with load order and that makes a lot of sense to me. Given that, I don't think we need anything to block updates to the strategies list. I'm happy here, @rarruda might spot something that I've missed though
> 
> In terms of the conversation about backwards compatibility I definitely agree, we've known for a while that they way we loaded custom strategies wasn't what we wanted it to be. That being said I have two worries here, firstly there are users in the wild that are currently using that approach and humans are terribly bad at reading patch notes, even in major releases. Secondly, adding breaking changes is going to lock people out of updates here until they can update to make use of the approach in this PR. I want to say that ideally we should provide backwards compatibly for a period of time, which should give us time to warn folks in our docs and give them breathing room to get this updated smoothly. Definitely agree that we should move towards removing the old approach soon. Is it a horrible amount of work to support both for a period of time?

No, it is not a lot of work to provide backward compatibility with both cases. I will extend my PR by end of the week. If you don't mind - I will leave `Strategies` class, cause it will allow centralizing all future and backward compatible code related to managing list of strategies.

volodymyr-mykhailyk:
@sighphyre, @rarruda Added backward compatibility with previous registration methods for custom strategies, including deprecation warning. Please let me know if there is anything else I can do or any other changes required per my questions above.

lib/unleash/strategies.rb
```diff
@@ -64,5 +70,11 @@ def register_base_strategies
       Unleash::Strategy::RemoteAddress,
       Unleash::Strategy::UserWithId
     ].freeze
+
+    def warn_deprecated_registration(strategy, method)
+      warn "[DEPRECATED] Registering custom Unleash strategy by #{method} is deprecated.
```
rarruda, line 75: We currently have a logger instance where all output goes to. (in case someone would like to suppress warnings, for whatever reason). Would be nice to keep using it.

rarruda:
> > @volodymyr-mykhailyk Thanks for the detailed response, that makes a lot of sense! To address a few questions here:
> > > Question: If you think it is still too much abstraction - let me know. Will happily remove wrapper.
> > 
> > 
> > My vote here is to use a hash over the class. It feels like more abstraction than we currently need (if we can tackle the backwards compatibility conversation without it).

> No, it is not a lot of work to provide backward compatibility with both cases. I will extend my PR by end of the week. If you don't mind - I will leave `Strategies` class, cause it will allow centralizing all future and backward compatible code related to managing list of strategies. Also, this class can be used to provide Hash like interface.

I would have voted for removing the `Strategies` class wrapper, and keep it as a `Hash`. :see_no_evil:

lib/unleash/strategies.rb
volodymyr-mykhailyk: For deprecation warnings, it is common to use the Kerner warning mechanism as it allows engineers to see warnings regardless of local logger configuration. Example:
- [Gem native way to notify deprecation](https://docs.ruby-lang.org/en/2.1.0/Gem/Deprecate.html)
- [Rails default deprecation behavior](https://apidock.com/rails/ActiveSupport/Deprecation/Behavior/behavior)

In this case, I decided to simply use `Kernel.warn` as Gem related functionality does not allow creating a custom message for the warning.

volodymyr-mykhailyk:
> > > @volodymyr-mykhailyk Thanks for the detailed response, that makes a lot of sense! To address a few questions here:
> > > > Question: If you think it is still too much abstraction - let me know. Will happily remove wrapper.
> > > 
> > > 
> > > My vote here is to use a hash over the class. It feels like more abstraction than we currently need (if we can tackle the backwards compatibility conversation without it).
> 
> > No, it is not a lot of work to provide backward compatibility with both cases. I will extend my PR by end of the week. If you don't mind - I will leave `Strategies` class, cause it will allow centralizing all future and backward compatible code related to managing list of strategies. Also, this class can be used to provide Hash like interface.
> 
> I would have voted for removing the `Strategies` class wrapper, and keep it as a `Hash`. 🙈
> 
> Also, sorry for being extremely slow on providing feedback.

I can remove `Strategies` class, but deprecation warnings will require another proxy class between `Unleash::STRATEGIES` and `Unleash.configuration.strategies`.

Should I give it a try or simply remove all the abstraction together with warnings?

sighphyre:
@volodymyr-mykhailyk We've chatted a little bit about that, I think the general feeling is that we're happy. When we pull out the legacy strategy implementation we'll refactor that if we feel it's necessary. This is a great piece of work, thank you so much for this and for your patience. There's a little niggle on our build at the moment, if you don't mind I'm going to rebase this and then merge once the tests are green?

volodymyr-mykhailyk:
@sighphyre Thanks. The code is yours. Do whatever is necessary to get it merged. 

I was annoyed with GitHub actions failing as well, so tried to help in separate PR #100

volodymyr-mykhailyk:
@sighphyre Anything still blocking this change? Can I help to get it merged in any way? Thanks

volodymyr-mykhailyk:
@gardleopard thanks for follow up. When are you planning to release new version?

sighphyre:
@volodymyr-mykhailyk We should get this out first thing Monday morning

PR #97, opened 2022-05-03
Ensure backup file is properly closed.

imaximix:
## Reason
We introduced this excellent gem, in a project of ours and noticed these strange errors in the logs:
```
ToggleFetcher ERROR] : thread ToggleFetcher threw exception NoMethodError  (1/Infinity): 'undefined method `close' for <Rake::FileTask  => []>:Rake::FileTask

        file&.close if defined?(file)
            ^^^^^^^
```
It turned out that the file variable was out of scope for the ensure block in the `Unleash::ToggleFetcher#save!` and in this case it was referring to a DSL method in the rake gem.

https://github.com/ruby/rake/blob/881416e322d597e47e62935ef6a9dbaed934fbfc/lib/rake/dsl_definition.rb#L76

This revealed some problems when closing the file in the ToggleFetcher.

## About the changes
Changes the way the file is closed in `Unleash::ToggleFetcher#save!`.
The ensure block didn't previously close the file because the variable was out of scope.
We also removed the explicit call to `Mutex#unlock` because using `Mutex#synchronize` unlocks when exiting the given block. 
https://github.com/ruby/ruby/blob/42a0bed351979cb4a59c641fa5f03e49609561fd/thread_sync.c#L600

We also added some defaults in spec_helper.rb to make it easier to write tests.

### DIFF
--- a/spec/unleash/toggle_fetcher_spec.rb
+++ b/spec/unleash/toggle_fetcher_spec.rb
@@ -1,0 +1,55 @@
+require 'spec_helper'
+
+RSpec.describe Unleash::ToggleFetcher do
+  subject(:toggle_fetcher) { Unleash::ToggleFetcher.new }
+
+  describe '#save!' do
+    before do
+      Unleash.configure do |config|
+        config.url      = 'http://toggle-fetcher-test-url/'
+        config.app_name = 'toggle-fetcher-my-test-app'
+      end
+
+      fetcher_features = {
+        "version": 1,
+        "features": [
+          {
+            "name": "Feature.A",
+            "description": "Enabled toggle",
+            "enabled": true,
+            "strategies": [{ "name": "toggle-fetcher" }]
+          }
+        ]
+      }
+
+      WebMock.stub_request(:get, "http://toggle-fetcher-test-url/client/features")
+        .to_return(status: 200,
+                   body: fetcher_features.to_json,
+                   headers: {})
+
+      Unleash.logger = Unleash.configuration.logger
+    end
+
+    context 'when toggle_cache generation fails' do
+      before do
+        allow(toggle_fetcher).to receive(:toggle_cache).and_raise(StandardError)
+      end
+
+      it 'swallows the error' do
+        expect { toggle_fetcher.save! }.not_to raise_error
+      end
+    end
+
+    context 'when toggle_cache with content is saved' do
+      before do
+        toggle_fetcher.toggle_cache = { features: [] }
+      end
+
+      it 'creates a file with toggle_cache in JSON' do
+        toggle_fetcher.save!
+        expect(File.exist?(Unleash.configuration.backup_file)).to eq(true)
+        expect(File.read(Unleash.configuration.backup_file)).to eq('{"features":[]}')
+      end
+    end
+  end
+end
--- a/lib/unleash/toggle_fetcher.rb
+++ b/lib/unleash/toggle_fetcher.rb
@@ -65,24 +65,20 @@ module Unleash
 
     def save!
       Unleash.logger.debug "Will save toggles to disk now"
-      begin
-        backup_file = Unleash.configuration.backup_file
-        backup_file_tmp = "#{backup_file}.tmp"
 
-        self.toggle_lock.synchronize do
-          file = File.open(backup_file_tmp, "w")
+      backup_file = Unleash.configuration.backup_file
+      backup_file_tmp = "#{backup_file}.tmp"
+
+      self.toggle_lock.synchronize do
+        File.open(backup_file_tmp, "w") do |file|
           file.write(self.toggle_cache.to_json)
-          file.close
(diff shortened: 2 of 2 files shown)


imaximix:
Updated the description.

spec/unleash/toggle_fetcher_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      end
+
+      it 'closes the file properly' do
+        toggle_fetcher.save!
+
+        backup_file = Unleash.configuration.backup_file
+        backup_file_tmp = "#{backup_file}.tmp"
+
+        backup_files = ObjectSpace.each_object(File).select { |f| f.path == backup_file_tmp }
+
```
sighphyre, line 19: There's a tiny bit of trailing white space here that Rubocop seems to not like

```suggestion

```

sighphyre (changes requested):
Thanks for this! I think this looks good. There's a tiny issue here with white space and rubocop, there's a suggestion to fix that. Agree on the locking scheme, I want to say this should be released cleanly from the block regardless of whether or not an error is raised here

spec/spec_helper.rb
```diff
@@ -21,6 +21,27 @@
...
+      "description": "Enabled toggle",
+      "enabled": true,
+      "strategies": [{ "name": "default" }]
+    }
+  ]
+}
+
+WebMock.stub_request(:get, "http://default-test-url/client/features").to_return(status: 200, body: default_features.to_json, headers: {})
+
+Unleash.logger = Unleash.configuration.logger
```
rarruda, line 43: I would prefer to keep these in the the scope of `toggle_fetcher_spec.rb` for now, as it is only used there, and can be confusing for others later when looking at the test code of what is relevant to which _spec tester.

Eventually it would make sense to have some of most common settings in a common configuration file, but I see that as a separate PR.

rarruda (commented):
I love it!

The code is simpler and cleaner now, as well as with a nice fix. :)

I would just like to keep the default settings for the tests within their own `_spec.rb` to make it easy to know what settings apply in each context. This can be re-evaluated, as I do see pro's in the suggested approach, but IMHO it should come in a separate PR.

spec/spec_helper.rb
anakinj: Ok, makes sense. Moved the setup to the spec itself.

spec/unleash/toggle_fetcher_spec.rb
anakinj: Sharp eyes here. Thanks for the comment.

sighphyre (approved):

FredrikOseberg:
[Github PR: unleash-client-ruby - Ensure backup file is properly closed.](https://trello.com/c/6bwmh2kX/1138-github-pr-unleash-client-ruby-ensure-backup-file-is-properly-closed)

spec/unleash/toggle_fetcher_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+          {
+            "name": "Feature.A",
+            "description": "Enabled toggle",
+            "enabled": true,
+            "strategies": [{ "name": "toggle-fetcher" }]
+          }
+        ]
+      }
+
+      WebMock.stub_request(:get, "http://toggle-fetcher-test-url/client/features").to_return(status: 200,
```
rarruda, line 23: Nit picking (so, not important): Could I recommend braking this into two lines? (for styling/legibility sake)

(.to_return) in a separate line.

rarruda (commented):
Looks great! Left just a super minor nit pick comment.

sighphyre:
@imaximix This looks great, we're happy with the changes. I'd really like to merge this but we've had some trouble with our build, I've updated the main branch here with the fix (mostly just lockfile). Would you mind updating your fork with latest changes from main so we can get the build to run correctly? Thank you!

Ben3eeE:
@sighphyre I rebased from upstream/main. Let's see if that fixes the build issues.

sighphyre:
@imaximix Thank you! That seems to have got the build happy again. It does look like one of the tests you've added doesn't play nice with JRuby though. Would you mind taking a look? Ideally, I'll like to not have to fiddle with the build config to get that to work for a single test but I'm open if there's no clean or rational way to do what you need to do there

Ben3eeE:
@sighphyre We changed the tests to more explicitly test the functionality of the save method instead of testing that the file was closed correctly. The test for the file being closed was useful during development but it doesn't add much value to test this in production because it just verifies that Ruby blocks are working correctly.

rarruda:
Would it be possible to rebase this branch from the current main ?  If you prefer to merge from `main` into your branch that's also fine. 

(As the build process should now be working again) And it would be nice to have a green build before merging.

rarruda (approved):

sighphyre:
@Ben3eeE @imaximix You folks are amazing, thank you so much for this. I''m all happy, going to merge this now

PR #99, opened 2022-05-13
fix: Fix rubocop build

sighphyre:
Previously the project didn't include Gemfile.lock, however the github actions are configured to install the version of Rubocop found in the Gemfile.lock. Backtracking previous action runs it can be seen that this has been quietly failing and installing the latest version of Rubocop for a long time. That's worked very well in the past, but now Rubocop has dropped support for Ruby 2.5 in the latest version, which is breaking the build.

I feel a lockfile is a good thing to have and should solve this problem, however, the same thing can be achieved by pinning the version of Rubocop in the action yaml


thomasheartman (approved):
Nice! Seems very sensible to me 😄 I'll approve it here, but it might be worth waiting for @rarruda before actually merging it 💁🏼

sighphyre:
@thomasheartman Thank you! And good point - handled over a slack conversation

PR #100, opened 2022-05-18
Fix GitHub actions

volodymyr-mykhailyk:
Minor modification to github actions to make them compatible with forked repos and ruby 2.5

Unfortunately, I failed to find suitable version of rubocop via `reviewdog` action. That's why rubocop is disabled on old rubies. Technically - running rubocop on all versions is redundant as it's result depends on code, not runtime. 

As a suggestion, I can recommend splitting this into two jobs. One for static test on single ruby, another for matrix rspec testing.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -2,6 +2,8 @@ name: CI
 
 on:
   push:
+    branches:
+      - master
   pull_request:
 
 jobs:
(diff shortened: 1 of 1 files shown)


nunogois:
Hi @volodymyr-mykhailyk - Thanks for opening this PR! We've seen this and will get back with more information soon.

sighphyre:
@volodymyr-mykhailyk @nunogois Apologies for not following up on this one, folks. We've actually already fixed this in a separate PR so I'm going to close this one

PR #101, opened 2022-05-20
Fix the rubocop step in CI

Ben3eeE:
## About the changes

We fixed RuboCop to be consistent on the local environment and CI by pinning it in the gemspec file and removing the flaky GitHub action with a new step that is only run once for a PR.

The version is pinned to < 1.0.0 because using a more recent version will result in failures and upgrading RuboCop can be done in a separate PR if this is merged.

This also removes the Gemfile.lock which was added recently in https://github.com/Unleash/unleash-client-ruby/pull/99.
Adding the Gemfile.lock to a gem fails to ensure that the dependency constraints in the gemspec are honored so the tests don't run with the latest version that someone installing the gem would use.


sighphyre (approved):
Thank you so much for this! This is a big improvement

PR #104, opened 2022-06-07
Require spec_helper in the .rspec file

Ben3eeE:
## About the changes
Require the spec_helper file globally when running tests so it's not required to be included first in every test file. This would be a first step towards implementing some defaults as discussed in https://github.com/Unleash/unleash-client-ruby/pull/97#discussion_r865787364

It also makes it easier to contribute changes because there is no need to know which files are required to be imported in the test context.

## Discussion points
Not every project uses this approach and it's more of a preference which method is used.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,2 +1,3 @@
 --format documentation
 --color
+--require 'spec_helper'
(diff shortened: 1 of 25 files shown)


sighphyre (approved):
Nice, LGTM, thanks for this!

rarruda (approved):

PR #105, opened 2022-06-07
Update RuboCop version

Ben3eeE:
## About the changes
Updates the RuboCop version to version 1.28.2. RuboCop 1.28 is the latest version to still support Ruby 2.5.
This is the version that was intended to be used but the build was broken and after it was fixed in https://github.com/Unleash/unleash-client-ruby/pull/101 we can now update the version and fix the RuboCop errors.

### Important files
* .rubocop.yml
* lib/unleash/variant_definition.rb

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -34,6 +34,9 @@ Style/StringLiterals:
 Style/RedundantSelf:
   Enabled: false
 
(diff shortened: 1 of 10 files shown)


sighphyre (approved):
Thanks for doing this! Looks good to me

PR #106, opened 2022-06-07
Refactor toggle fetcher read

anakinj:
## About the changes
Added tests and simplified reading of the backup_file in `ToggleFetcher#read!`. Continues the work stared in #97

### DIFF
--- a/spec/unleash/toggle_fetcher_spec.rb
+++ b/spec/unleash/toggle_fetcher_spec.rb
@@ -2,10 +2,12 @@ require 'spec_helper'
 
 RSpec.describe Unleash::ToggleFetcher do
   subject(:toggle_fetcher) { Unleash::ToggleFetcher.new }
+
   before do
     Unleash.configure do |config|
       config.url      = 'http://toggle-fetcher-test-url/'
       config.app_name = 'toggle-fetcher-my-test-app'
+      config.disable_client = false # Some test is changing the process state for this one
     end
 
     fetcher_features = {
@@ -26,6 +28,11 @@ RSpec.describe Unleash::ToggleFetcher do
                  headers: {})
 
     Unleash.logger = Unleash.configuration.logger
+  end
+
+  after do
+    WebMock.reset!
+    File.delete(Unleash.configuration.backup_file) if File.exist?(Unleash.configuration.backup_file)
   end
 
   describe '#save!' do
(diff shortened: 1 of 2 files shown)


spec/unleash/toggle_fetcher_spec.rb
```diff
@@ -52,4 +58,31 @@
       end
     end
   end
+
+  describe '.new' do
+    context 'when there are problems fetching toggles' do
+      before do
+        cache_creator = described_class.new
```
rarruda, line 65:
```suggestion
        # manually create a stub cache on disk
        cache_creator = described_class.new
```

spec/unleash/toggle_fetcher_spec.rb
```diff
@@ -52,4 +58,31 @@
...
+        cache_creator.toggle_cache = { features: [] }
+        cache_creator.save!
+
+        WebMock.stub_request(:get, "http://toggle-fetcher-test-url/client/features").to_return(status: 500)
+      end
+
+      it 'reads the backup file for values' do
+        expect(toggle_fetcher.toggle_cache).to eq("features" => [])
```
rarruda, line 73: Would have been nice to test that `File.read` was called, but I guess that's a bunch of work, when we are pretty sure already that it works here.

rarruda, line 73: Nitpick: Symbols vs Strings again 😄 
```suggestion
        expect(toggle_fetcher.toggle_cache).to eq(features: [])
```

rarruda (approved):
Looks great. The refactored version is clearly simpler and better.

I added a couple of suggestions, but they are pretty much nit-picking/bike shedding.

spec/unleash/toggle_fetcher_spec.rb
anakinj: `JSON.parse` uses strings as keys unless the `symbolize_names` parameter is given to it. Using symbols as keys here would fail the expectation.

anakinj: I would argue that `File.read` or the previous way of reading (File.open+file.read) is maybe more an implementation detail, the test would not care how the data ends up in the correct place.

rarruda: Good point!

PR #107, opened 2022-06-07
Add dynamic headers support

### LINKED ISSUE #83: feat: add support for dynamic headers
for reference: 

https://github.com/Unleash/unleash-client-python/issues/104
https://github.com/Unleash/unleash-client-node/issues/151
https://github.com/Unleash/unleash-client-node/pull/152
https://github.com/Unleash/unleash-client-java#dynamic-custom-http-headers

Ben3eeE:
## About the changes
Add support for dynamic `custom_http_headers` by allowing passing a Proc/Lambda as the configuration.

We also removed and unnecessary duplication from the `http_headers` method.

Closes #83 

## Discussion points
We removed a redundant method `ensure_valid_opts` that was called in the constructor. This method was also out of sync with the `Configuration#validate!` by allowing nil. We ensured that `Configuration#validate!` was called in the places where it's needed before removing this method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,8 +67,13 @@ If you need custom http headers that change during the lifetime of the client, a
 ```ruby
 Unleash.configure do |config|
   config.app_name            = 'my_ruby_app'
-  config.url                 = 'http://unleash.herokuapp.com/api'
-  config.custom_http_headers =  -> { {'Authorization': '<API token>'} }
+  config.url                 = 'https://unleash.herokuapp.com/api'
+  config.custom_http_headers =  proc do
+    {
+      'Authorization': '<API token>',
+      'X-Client-Request-Time': Time.now.iso8601
+    }
+  end
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -61,6 +61,17 @@ or instantiate the client with the valid configuration:
 UNLEASH = Unleash::Client.new(url: 'http://unleash.herokuapp.com/api', app_name: 'my_ruby_app', custom_http_headers: {'Authorization': '<API token>'})
 ```
 
+## Dynamic custom HTTP headers
+If you need custom http headers that change during the lifetime of the client, a the `custom_http_headers` can be given as a `Proc`.
+
+```ruby
+Unleash.configure do |config|
+  config.app_name            = 'my_ruby_app'
+  config.url                 = 'http://unleash.herokuapp.com/api'
+  config.custom_http_headers =  -> { {'Authorization': '<API token>'} }
+end
+```
+
 #### List of Arguments
 
 Argument | Description | Required? |  Type |  Default Value|
@@ -74,7 +85,7 @@ Argument | Description | Required? |  Type |  Default Value|
 `metrics_interval` | How often the unleash client should send metrics to server. | N | Integer | 60 |
 `disable_client` | Disables all communication with the Unleash server, effectively taking it *offline*. If set, `is_enabled?` will always answer with the `default_value` and configuration validation is skipped. Defeats the entire purpose of using unleash, but can be useful in when running tests. | N | Boolean | `false` |
 `disable_metrics` | Disables sending metrics to Unleash server. | N | Boolean | `false` |
-`custom_http_headers` | Custom headers to send to Unleash. As of Unleash v4.0.0, the `Authorization` header is required. For example: `{'Authorization': '<API token>'}` | N | Hash | {} |
+`custom_http_headers` | Custom headers to send to Unleash. As of Unleash v4.0.0, the `Authorization` header is required. For example: `{'Authorization': '<API token>'}` | N | Hash/Proc | {} |
 `timeout` | How long to wait for the connection to be established or wait in reading state (open_timeout/read_timeout) | N | Integer | 30 |
 `retry_limit` | How many consecutive failures in connecting to the Unleash server are allowed before giving up. The default is to retry indefinitely. | N | Float::INFINITY | 5 |
 `backup_file` | Filename to store the last known state from the Unleash server. Best to not change this from the default. | N | String | `Dir.tmpdir + "/unleash-#{app_name}-repo.json` |
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ UNLEASH = Unleash::Client.new(url: 'http://unleash.herokuapp.com/api', app_name:
 ```
 
 ## Dynamic custom HTTP headers
-If you need custom http headers that change during the lifetime of the client, a the `custom_http_headers` can be given as a `Proc`.
+If you need custom HTTP headers that change during the lifetime of the client, the `custom_http_headers` can be given as a `Proc`.
 
 ```ruby
 Unleash.configure do |config|
--- a/lib/unleash/configuration.rb
+++ b/lib/unleash/configuration.rb
@@ -52,7 +52,7 @@ module Unleash
       {
         'UNLEASH-INSTANCEID' => self.instance_id,
         'UNLEASH-APPNAME' => self.app_name
-      }.merge(generate_custom_http_headers)
+      }.merge!(generate_custom_http_headers)
     end
 
     def fetch_toggles_uri
--- a/lib/unleash/configuration.rb
+++ b/lib/unleash/configuration.rb
@@ -40,7 +40,7 @@ module Unleash
 
       raise ArgumentError, "URL and app_name are required parameters." if self.app_name.nil? || self.url.nil?
       unless self.custom_http_headers.is_a?(Hash) || self.custom_http_headers.respond_to?(:call)
-        raise ArgumentError, "custom_http_headers must be a hash."
+        raise ArgumentError, "custom_http_headers must be a Hash or a Proc."
       end
     end
 
--- a/lib/unleash/configuration.rb
+++ b/lib/unleash/configuration.rb
@@ -23,6 +23,7 @@ module Unleash
       :bootstrap_config
 
     def initialize(opts = {})
+      validate_custom_http_headers!(opts[:custom_http_headers]) if opts.has_key?(:custom_http_headers)
       set_defaults
 
       initialize_default_logger if opts[:logger].nil?
@@ -39,9 +40,8 @@ module Unleash
       return if self.disable_client
 
       raise ArgumentError, "URL and app_name are required parameters." if self.app_name.nil? || self.url.nil?
-      unless self.custom_http_headers.is_a?(Hash) || self.custom_http_headers.respond_to?(:call)
-        raise ArgumentError, "custom_http_headers must be a Hash or a Proc."
-      end
+
+      validate_custom_http_headers!(self.custom_http_headers)
     end
 
     def refresh_backup_file!
@@ -113,12 +113,16 @@ module Unleash
       self
     end
 
+    def validate_custom_http_headers!(custom_http_headers)
+      return if custom_http_headers.is_a?(Hash) || custom_http_headers.respond_to?(:call)
+
+      raise ArgumentError, "custom_http_headers must be a Hash or a Proc."
+    end
+
     def generate_custom_http_headers
-      if self.custom_http_headers.respond_to?(:call)
-        self.custom_http_headers.call
-      else
-        self.custom_http_headers
-      end
+      return self.custom_http_headers.call if self.custom_http_headers.respond_to?(:call)
+
+      self.custom_http_headers
     end
 
     def set_option(opt, val)
--- a/lib/unleash/configuration.rb
+++ b/lib/unleash/configuration.rb
@@ -23,7 +23,6 @@ module Unleash
       :bootstrap_config
 
     def initialize(opts = {})
-      ensure_valid_opts(opts)
       set_defaults
 
       initialize_default_logger if opts[:logger].nil?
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -61,6 +61,17 @@ or instantiate the client with the valid configuration:
 UNLEASH = Unleash::Client.new(url: 'http://unleash.herokuapp.com/api', app_name: 'my_ruby_app', custom_http_headers: {'Authorization': '<API token>'})
 ```
 
+## Dynamic custom HTTP headers
+If you need custom http headers that change during the lifetime of the client, a the `custom_http_headers` can be given as a `Proc`.
+
+```ruby
+Unleash.configure do |config|
+  config.app_name            = 'my_ruby_app'
+  config.url                 = 'http://unleash.herokuapp.com/api'
```
sighphyre, line 70:
```suggestion
  config.url                 = 'https://unleash.herokuapp.com/api'
```

Ben3eeE: This was copied from the example above, should we change both places?

spec/unleash/configuration_spec.rb
```diff
@@ -107,23 +107,35 @@
       Unleash.configure do |config|
         config.url      = 'http://test-url/'
         config.app_name = 'my-test-app'
-        config.custom_http_headers = { 'X-API-KEY': '123' }
+        config.custom_http_headers = { 'X-API-KEY' => '123' }
```
rarruda, line 110: Would prefer to keep any previous test unchanged, to the degree that it is possible. This way we can be more secure that there are no known side effects on code changes.

In a separate note, I do like using symbols better than strings for the keys, but that's secondary. Both should work.

spec/unleash/configuration_spec.rb
```diff
@@ -107,23 +107,35 @@
...
-        custom_http_headers: { 'X-API-KEY': '123' }
+        custom_http_headers: { 'X-API-KEY' => '123' }
       )
 
       expect{ config.validate! }.not_to raise_error
-      expect(config.custom_http_headers).to include({ 'X-API-KEY': '123' })
-      expect(config.http_headers).to include({ 'UNLEASH-APPNAME' => 'test-app' })
-      expect(config.http_headers).to include('UNLEASH-INSTANCEID')
+      expect(config.custom_http_headers).to eq({ 'X-API-KEY' => '123' })
+      expect(config.http_headers).to eq('X-API-KEY' => '123', 'UNLEASH-APPNAME' => 'test-app', 'UNLEASH-INSTANCEID' => config.instance_id)
```
rarruda, line 126: (Bike-shedding) For readability would prefer to have to have the hash broken into multiple lines:
```suggestion
      expect(config.http_headers).to eq({
        'X-API-KEY' => '123',
        'UNLEASH-APPNAME' => 'test-app',
        'UNLEASH-INSTANCEID' => config.instance_id
       })
```

README.md
```diff
@@ -61,6 +61,17 @@ or instantiate the client with the valid configuration:
...
+  config.custom_http_headers =  -> { {'Authorization': '<API token>'} }
```
rarruda, line 71: Would prefer to give a bit more explicit example:

```suggestion
  config.custom_http_headers =  Proc.new{ 

{'Authorization': '<API token>', } 
  }
```

lib/unleash/configuration.rb
```diff
@@ -23,7 +23,6 @@ class Configuration
       :bootstrap_config
 
     def initialize(opts = {})
-      ensure_valid_opts(opts)
```
rarruda, line 26: I would prefer to keep the `ensure_valid_opts()` method for now.

We can call it from the `validate!` method, to avoid code duplication, if that's a significant concern. Otherwise we can de-duplicate that in a separate PR.

lib/unleash/configuration.rb
```diff
@@ -118,6 +113,14 @@ def merge(opts)
       self
     end
 
+    def generate_custom_http_headers
+      if self.custom_http_headers.respond_to?(:call)
+        self.custom_http_headers.call
+      else
+        self.custom_http_headers
+      end
```
rarruda, line 121: Would suggest this alternative syntax to make it more ruby idiomatic and easier to read:
```suggestion
      return self.custom_http_headers.call if self.custom_http_headers.respond_to?(:call)

      self.custom_http_headers.dup
```

Note that we duplicate the value of the `custom_http_headers` Hash, to ensure that they are not unexpectedly changed later on after the initialization of the client.

spec/unleash/configuration_spec.rb
```diff
@@ -107,23 +107,35 @@
...
+      expect(Unleash.configuration.custom_http_headers).to eq({ 'X-API-KEY' => '123' })
+      expect(Unleash.configuration.http_headers).to include({ 'X-API-KEY' => '123' })
     end
 
     it "should allow hashes for custom_http_headers via new client" do
       config = Unleash::Configuration.new(
         url: 'https://testurl/api',
         app_name: 'test-app',
-        custom_http_headers: { 'X-API-KEY': '123' }
+        custom_http_headers: { 'X-API-KEY' => '123' }
```
rarruda, line 121: Same as above (strings vs symbols)

spec/unleash/configuration_spec.rb
```diff
@@ -107,23 +107,35 @@
...
+    it "should allow lambdas and procs for custom_https_headers via new client" do
+      config = Unleash::Configuration.new(
+        url: 'https://testurl/api',
+        app_name: 'test-app',
+        custom_http_headers: -> { { 'X-API-KEY' => '123' } }
+      )
+
+      expect{ config.validate! }.not_to raise_error
+      expect(config.custom_http_headers).to be_a(Proc)
+      expect(config.http_headers).to eq('X-API-KEY' => '123', 'UNLEASH-APPNAME' => 'test-app', 'UNLEASH-INSTANCEID' => config.instance_id)
```
rarruda, line 138: (same bike-shedding comment regarding breaking up hash into multiple lines).

spec/unleash/configuration_spec.rb
```diff
@@ -133,7 +145,7 @@
           config.app_name = 'my-test-app'
           config.custom_http_headers = 123.456
         end
-      end.to raise_error(ArgumentError)
+      end.to raise_error(ArgumentError, "custom_http_headers must be a Hash or a Proc.")
```
rarruda, line 148: :heart:

spec/unleash/configuration_spec.rb
```diff
@@ -107,23 +107,35 @@
...
-      expect(config.http_headers).to include('UNLEASH-INSTANCEID')
+      expect(config.custom_http_headers).to eq({ 'X-API-KEY' => '123' })
+      expect(config.http_headers).to eq('X-API-KEY' => '123', 'UNLEASH-APPNAME' => 'test-app', 'UNLEASH-INSTANCEID' => config.instance_id)
+    end
+
+    it "should allow lambdas and procs for custom_https_headers via new client" do
+      config = Unleash::Configuration.new(
+        url: 'https://testurl/api',
+        app_name: 'test-app',
+        custom_http_headers: -> { { 'X-API-KEY' => '123' } }
```
rarruda, line 133: Would like to have a non-static `Proc` getting called in the test. Possibly calling a method/class that is mocked, so we can then test that the mocked class has been actually been called.

README.md
sighphyre: If you don't mind! It's a very small thing but its nice to keep this kind of stuff updated when we are touching the code anyway

rarruda (commented):
This is great work and quite promising! Left some feedback, some of it nit picking, and only a few things bit more important.

rarruda:
Sorry, looks like you got an early partial review. Will resolve the issues there so you only have one place to look at.

spec/unleash/configuration_spec.rb
sighphyre: This is actually a great comment and I think if this test has to change to make this PR work then we probably need to talk a little bit more about details there. Mostly because there's already config in the wild and we should strive to not break that contract if possible

lib/unleash/configuration.rb
```diff
@@ -40,7 +39,9 @@ def validate!
       return if self.disable_client
 
       raise ArgumentError, "URL and app_name are required parameters." if self.app_name.nil? || self.url.nil?
-      raise ArgumentError, "custom_http_headers must be a hash." unless self.custom_http_headers.is_a?(Hash)
+      unless self.custom_http_headers.is_a?(Hash) || self.custom_http_headers.respond_to?(:call)
+        raise ArgumentError, "custom_http_headers must be a Hash or a Proc."
+      end
```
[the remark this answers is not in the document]
rarruda: :cry: 

Good point. Maybe we can rename `ensure_valid_opts(opts)` into `valid_custom_http_headers?(custom_http_headers)` and move the logic there, so the line here becomes shorter?

```ruby
   def valid_custom_http_headers?(custom_http_headers)
      custom_http_headers.is_a?(Hash) || custom_http_headers.respond_to?(:call)
   end
```

It could then be called from both here, and `ensure_valid_opts()` is originally called from as `ensure_valid_opts(self.custom_http_headers)`

lib/unleash/configuration.rb
Ben3eeE: The `Configuration#http_headers` method already creates two new instances of a Hash so it's not needed to duplicate. And if we add the duplication here it's just the Hash that gets passed to the merge function that is duplicated. Not the result of the `Configuration#https_headers` method.

Ben3eeE: We would suggest changing the `merge` to be `merge!` in `Configuration#http_headers` to avoid creating two Hashes there and just merge it into the hash that is created in the beginning of the method.

spec/unleash/configuration_spec.rb
Ben3eeE: We changed this because the Hash that is created in `Configuration#http_headers` is using strings as keys. All examples and tests use symbol notation which creates a mix of strings and symbols in the resulting object.

Mixing strings and symbols as keys makes the result confusing. So we wanted to avoid using symbols as keys in the existing tests.

```ruby
config = Unleash::Configuration.new(app_name: 'static', custom_http_headers: { 'UNLEASH-APPNAME': 'dynamic' })

config.http_headers
# => {"UNLEASH-INSTANCEID"=>"0c52c178-dafd-4b56-b0ce-424ce9b58848", "UNLEASH-APPNAME"=>"static", :"UNLEASH-APPNAME"=>"dynamic"}

config.http_headers["UNLEASH-APPNAME"]
# => "static"

config.http_headers[:"UNLEASH-APPNAME"]
# => "dynamic"

config.http_headers.keys.map(&:to_s)
# => ["UNLEASH-INSTANCEID", "UNLEASH-APPNAME", "UNLEASH-APPNAME"]
```

Our recommendation would be to change the tests and examples to use the same notation. We can keep the old tests to ensure that existing configurations still work.

We can open a separate PR to change this though. What do you think?

lib/unleash/configuration.rb
```diff
@@ -23,7 +23,7 @@ class Configuration
       :bootstrap_config
 
     def initialize(opts = {})
-      ensure_valid_opts(opts)
+      validate_custom_http_headers!(opts[:custom_http_headers]) if opts.has_key?(:custom_http_headers)
```
rarruda, line 26: Total nit-picking: This is super duper valid, and good code. And I might even like it better than my own suggestion.

But a couple of lines below I use a different notation, so I think it would be a bit nice to be consistent with that. Feel free to ignore this feedback if you disagree.

```suggestion
      validate_custom_http_headers!(opts[:custom_http_headers]) unless opts[:custom_http_headers].nil?
```

spec/unleash/configuration_spec.rb
```diff
@@ -126,14 +126,38 @@
...
+      expect{ config.validate! }.not_to raise_error
+      expect(config.custom_http_headers).to be_a(Proc)
+      expect(config.http_headers).to eq(
+        {
+          'X-API-KEY' => '123',
+          'UNLEASH-APPNAME' => 'test-app',
+          'UNLEASH-INSTANCEID' => config.instance_id
+        }
+      )
+      expect(custom_headers_proc).to have_received(:call)
```
rarruda, line 150: Nit picking again:

```suggestion
      expect(custom_headers_proc).to have_received(:call).twice
```
Or the correct number of times...

```suggestion
      expect(custom_headers_proc).to have_received(:call).exactly(2).times
```

lib/unleash/configuration.rb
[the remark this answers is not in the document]
rarruda: Loved it!

lib/unleash/configuration.rb
rarruda: This was resolved as discussed at https://github.com/Unleash/unleash-client-ruby/pull/107#discussion_r892358994

lib/unleash/configuration.rb
rarruda: I agree that no duplication is needed in this case.

Feel free to use `merge!`, that does look more elegant.

spec/unleash/configuration_spec.rb
rarruda: Strings vs Symbols as keys is a known long discussion. Strings and Symbols will however both work interchangeably, and symbols have a very tiny performance edge in large hashes. (Which is not likely ever to be the case).

If one were to do something about, I think the better approach would be to convert the code and examples to use symbols as keys, instead of Strings. 

But to be honest, I think it is quite reasonable to leave things as is for now. However if you feel compelled to suggest otherwise, I agree we should do it in a separate PR.

rarruda (commented):
The PR is looking great! We should be able to merge it soon. (And sorry for taking so long to get back to you).

I left a couple of last comments, but they are pretty much nit-picking, and would like to hear your thoughts.

lib/unleash/configuration.rb
Ben3eeE: As someone who is just starting to learn Ruby the way it is now makes more sense at a quick glance, so I left it as it is for now. But I also agree that it should maybe change for consistency.

PR #110, opened 2022-06-21
fix: add test case for missing context field

ivarconr:
## About the changes
We should validate that the SDK is not crashing if the use does not specify the context field we have a constraint on.


rarruda:
This should now be fixed in https://github.com/Unleash/unleash-client-ruby/pull/116.

PR #114, opened 2022-09-08
feat: implement global segments

sighphyre:
This implements global segments as per issue #112. There is one major deviation from the way this is implemented in other SDKs, this doesn't use a lazy generator, this is because this SDK doesn't cache computation as aggressively as the other SDKs.

Most of the tests are done through the client specifications but there's also some failing tests here, these aren't caused by this PR, but by upgrading the client specs. There's another PR incoming to fix that.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Install dependencies
         run: bundle install
       - name: Download test cases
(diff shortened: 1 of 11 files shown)


sighphyre:
This seems to have a problem with coveralls here but the tests are passing locally

gardleopard (approved):
LGTM, tests are successful locally here as well.

PR #115, opened 2022-09-08
fix: support int context in set comparison

gardleopard:
This makes it possible to have int values in context and compare it to sets from unleash which are strings.


rarruda:
quick question: will the unleash server always provide values as strings? 

or do we have to think about the possibility of it sending ints in the json payload?

sighphyre:
@rarruda Yep they're always strings

rarruda (approved):

PR #116, opened 2022-09-08
Fix/client spec constraint tests

gardleopard:
This makes us pass the test cases for constraints in:
https://github.com/Unleash/client-specification/blob/main/specifications/13-constraint-operators.json

### DIFF
--- a/lib/unleash/constraint.rb
+++ b/lib/unleash/constraint.rb
@@ -42,17 +42,11 @@ module Unleash
     end
 
     def matches_context?(context)
-      Unleash.logger.debug "Unleash::Constraint matches_context? value: #{self.value} context.get_by_name(#{self.context_name})" \
-        " #{context.get_by_name(self.context_name)} "
       match = matches_constraint?(context)
       self.inverted ? !match : match
     rescue KeyError
       Unleash.logger.warn "Attemped to resolve a context key during constraint resolution: #{self.context_name} but it wasn't \
       found on the context"
-
-      # when the operator is NOT_IN and there is no data, return true. In all other cases the operator doesn't match.
-      return true if operator == :NOT_IN
-
       false
     end
 
@@ -99,6 +93,12 @@ module Unleash
         Unleash.logger.warn "Invalid constraint operator: #{self.operator}, this should be unreachable. Always returning false."
         false
       end
+
+      # when the operator is NOT_IN and there is no data, return true. In all other cases the operator doesn't match.
+      return self.operator == :NOT_IN unless context.has_property(self.context_name)
+
+      Unleash.logger.debug "Unleash::Constraint matches_context? value: #{self.value} context.get_by_name(#{self.context_name})" \
+      " #{context.get_by_name(self.context_name)} "
 
       v = self.value.dup
       context_value = context.get_by_name(self.context_name)
--- a/lib/unleash/constraint.rb
+++ b/lib/unleash/constraint.rb
@@ -92,8 +92,8 @@ module Unleash
 
     # This should be a private method but for some reason this fails on Ruby 2.5
     def validate_constraint_value_type(operator, value)
-      raise ArgumentError, "context_name is not an Array" if LIST_OPERATORS.include?(operator) && value.is_a?(String)
-      raise ArgumentError, "context_name is not a String" if !LIST_OPERATORS.include?(operator) && value.is_a?(Array)
+      Unleash.logger.info "value is not an Array" if LIST_OPERATORS.include?(operator) && value.is_a?(String)
+      Unleash.logger.info "value is not a String" if !LIST_OPERATORS.include?(operator) && value.is_a?(Array)
     end
 
     private
--- a/lib/unleash/constraint.rb
+++ b/lib/unleash/constraint.rb
@@ -95,7 +95,7 @@ module Unleash
       end
 
       # when the operator is NOT_IN and there is no data, return true. In all other cases the operator doesn't match.
-      return self.operator == :NOT_IN unless context.has_property(self.context_name)
+      return self.operator == :NOT_IN unless context.property?(self.context_name)
 
       Unleash.logger.debug "Unleash::Constraint matches_context? value: #{self.value} context.get_by_name(#{self.context_name})" \
       " #{context.get_by_name(self.context_name)} "
--- a/lib/unleash/constraint.rb
+++ b/lib/unleash/constraint.rb
@@ -42,6 +42,8 @@ module Unleash
     end
 
     def matches_context?(context)
+      Unleash.logger.debug "Unleash::Constraint matches_context? value: #{self.value} context.get_by_name(#{self.context_name})"
+
       match = matches_constraint?(context)
       self.inverted ? !match : match
     rescue KeyError
@@ -93,8 +95,6 @@ module Unleash
         Unleash.logger.warn "Invalid constraint operator: #{self.operator}, this should be unreachable. Always returning false."
         false
       end
-
-      Unleash.logger.debug "Unleash::Constraint matches_context? value: #{self.value} context.get_by_name(#{self.context_name})"
 
       # when the operator is NOT_IN and there is no data, return true. In all other cases the operator doesn't match.
       return self.operator == :NOT_IN unless context.include?(self.context_name)
--- a/lib/unleash/constraint.rb
+++ b/lib/unleash/constraint.rb
@@ -47,8 +47,8 @@ module Unleash
       match = matches_constraint?(context)
       self.inverted ? !match : match
     rescue KeyError
-      Unleash.logger.warn "Attempted to resolve a context key during constraint resolution: #{self.context_name} but it wasn't " \
-        "found on the context"
+      Unleash.logger.warn "Attemped to resolve a context key during constraint resolution: #{self.context_name} but it wasn't \
+      found on the context"
       false
     end
 
@@ -57,8 +57,8 @@ module Unleash
       val2 = DateTime.parse(val2)
       yield(val1, val2)
     rescue ArgumentError
-      Unleash.logger.warn "Unleash::ConstraintMatcher unable to parse either context_value (#{val1}) " \
-        "or constraint_value (#{val2}) into a date. Returning false!"
+      Unleash.logger.warn "Unleash::ConstraintMatcher unable to parse either context_value (#{val1}) \
+      or constraint_value (#{val2}) into a date. Returning false!"
       false
     end
 
@@ -67,8 +67,8 @@ module Unleash
       val2 = Float(val2)
       yield(val1, val2)
     rescue ArgumentError
-      Unleash.logger.warn "Unleash::ConstraintMatcher unable to parse either context_value (#{val1}) " \
-        "or constraint_value (#{val2}) into a number. Returning false!"
+      Unleash.logger.warn "Unleash::ConstraintMatcher unable to parse either context_value (#{val1}) \
+      or constraint_value (#{val2}) into a number. Returning false!"
       false
     end
 
@@ -77,8 +77,8 @@ module Unleash
       val2 = Gem::Version.new(val2)
       yield(val1, val2)
     rescue ArgumentError
-      Unleash.logger.warn "Unleash::ConstraintMatcher unable to parse either context_value (#{val1}) " \
-        "or constraint_value (#{val2}) into a version. Return false!"
+      Unleash.logger.warn "Unleash::ConstraintMatcher unable to parse either context_value (#{val1}) \
+      or constraint_value (#{val2}) into a version. Return false!"
       false
     end
 
--- a/lib/unleash/constraint.rb
+++ b/lib/unleash/constraint.rb
@@ -32,7 +32,7 @@ module Unleash
           falling back to FALLBACK_VALIDATOR which skips this constraint"
         operator = "FALLBACK_VALIDATOR"
       end
-      self.validate_constraint_value_type(operator.to_sym, value)
+      self.log_inconsistent_constraint_configuration(operator.to_sym, value)
 
       self.context_name = context_name
       self.operator = operator.to_sym
@@ -87,7 +87,7 @@ module Unleash
     end
 
     # This should be a private method but for some reason this fails on Ruby 2.5
-    def validate_constraint_value_type(operator, value)
+    def log_inconsistent_constraint_configuration(operator, value)
       Unleash.logger.warn "value is a String, operator is expecting an Array" if LIST_OPERATORS.include?(operator) && value.is_a?(String)
       Unleash.logger.warn "value is an Array, operator is expecting a String" if !LIST_OPERATORS.include?(operator) && value.is_a?(Array)
     end
--- a/lib/unleash/constraint.rb
+++ b/lib/unleash/constraint.rb
@@ -1,5 +1,4 @@
 require 'date'
-
 module Unleash
   class Constraint
     attr_accessor :context_name, :operator, :value, :inverted, :case_insensitive
@@ -19,15 +18,18 @@ module Unleash
       DATE_BEFORE: ->(context_v, constraint_v){ on_valid_date(constraint_v, context_v){ |x, y| (x > y) } },
       SEMVER_EQ: ->(context_v, constraint_v){ on_valid_version(constraint_v, context_v){ |x, y| (x == y) } },
       SEMVER_GT: ->(context_v, constraint_v){ on_valid_version(constraint_v, context_v){ |x, y| (x < y) } },
-      SEMVER_LT: ->(context_v, constraint_v){ on_valid_version(constraint_v, context_v){ |x, y| (x > y) } }
+      SEMVER_LT: ->(context_v, constraint_v){ on_valid_version(constraint_v, context_v){ |x, y| (x > y) } },
+      FALLBACK_VALIDATOR: ->(context_v, constraint_v){true}
     }.freeze
 
     LIST_OPERATORS = [:IN, :NOT_IN, :STR_STARTS_WITH, :STR_ENDS_WITH, :STR_CONTAINS].freeze
 
     def initialize(context_name, operator, value = [], inverted: false, case_insensitive: false)
       raise ArgumentError, "context_name is not a String" unless context_name.is_a?(String)
-      raise ArgumentError, "operator does not hold a valid value:" + OPERATORS.keys unless OPERATORS.include? operator.to_sym
-
+      unless OPERATORS.include? operator.to_sym
+        Unleash.logger.warn "Operator #{operator} is not a supported operator, falling back to FALLBACK_VALIDATOR which skips this constraint"
+        operator = "FALLBACK_VALIDATOR"
+      end
       self.validate_constraint_value_type(operator.to_sym, value)
 
       self.context_name = context_name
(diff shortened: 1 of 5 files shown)


sighphyre (commented):
This looks good to me but give me a bit of time to double check the client spec tests, this got crazy and I want to make sure we aren't crazy too

sighphyre:
Okay, never mind, looked over the client spec, this does look correct and the failing client spec test cases are because... well... nothing can't be in anything - i.e. this is isolated to the oddities of the IN/NOT_IN operators

The remaining failures are due to global constraints not being implemented and should be closed by #114

sighphyre (changes requested):
Still looks good but can we make Rubocop happy

spec/unleash/constraint_spec.rb
```diff
@@ -20,12 +20,6 @@
 
       constraint = Unleash::Constraint.new('env', 'IN', ['dev', 'pre'])
       expect(constraint.matches_context?(context)).to be true
-
-      constraint = Unleash::Constraint.new('env', 'NOT_IN', ['dev', 'pre'])
-      expect(constraint.matches_context?(context)).to be false
-
-      constraint = Unleash::Constraint.new('env', 'NOT_IN', ['pre', 'prod'])
-      expect(constraint.matches_context?(context)).to be true
```
rarruda, line 28: what's the rationale behind removing these tests?

spec/unleash/constraint_spec.rb
```diff
@@ -384,24 +388,21 @@
...
-
-      string_constraints = ['NUM_EQ', 'NUM_GT', 'NUM_GTE', 'NUM_LT', 'NUM_LTE',
-                            'DATE_AFTER', 'DATE_BEFORE', 'SEMVER_EQ', 'SEMVER_GT', 'SEMVER_LT']
-      string_constraints.each do |operator_name|
-        Unleash::Constraint.new('env', operator_name, '')
-        expect do
-          Unleash::Constraint.new('env', operator_name, [])
-        end.to raise_error
-      end
+    it 'handles malconfiguration of constraint operators' do
```
rarruda, line 391:
```suggestion
    it 'handles gracefully invalid constraint operators' do
```

lib/unleash/constraint.rb
```diff
@@ -45,7 +47,17 @@ def matches_context?(context)
...
+    end
+
+    def key_error_handler()
+      case operator
+      when "NOT_IN".to_sym
+        # when there is no input data present, nothing can't be in anything.
+        true
+      else
+        false
+      end
```
rarruda, line 60:
```suggestion

      # when the operator is NOT_IN and there is no data, return true. In all other cases the operator doesn't match.
      return true if operator == :NOT_IN
      false
```

rarruda: this could be even simpler by just having:
```suggestion
    (operator == :NOT_IN)
```

But I think that might be too concise and reduce legibility.

spec/unleash/constraint_spec.rb
```diff
@@ -48,6 +42,16 @@
       expect(constraint.matches_context?(context)).to be true
     end
 
+    it 'matches based on property not set by operator NOT_IN value' do
```
rarruda, line 45:
```suggestion
    it 'matches based on a value NOT_IN in empty context field' do
```

spec/unleash/constraint_spec.rb
```diff
@@ -384,24 +388,21 @@
...
-      end
-
-      string_constraints = ['NUM_EQ', 'NUM_GT', 'NUM_GTE', 'NUM_LT', 'NUM_LTE',
-                            'DATE_AFTER', 'DATE_BEFORE', 'SEMVER_EQ', 'SEMVER_GT', 'SEMVER_LT']
-      string_constraints.each do |operator_name|
-        Unleash::Constraint.new('env', operator_name, '')
-        expect do
-          Unleash::Constraint.new('env', operator_name, [])
-        end.to raise_error
-      end
```
rarruda, line 404: you can maybe keep the tests but only check that the log message (with the warning) is called?

That is unexpected a different kind of unexpected behavior that would be nice to still have a test for. WDYT?

rarruda (commented):
left some suggestions.

but as @sighphyre commented, rubocop doesn't seem quite happy yet.

spec/unleash/constraint_spec.rb
sighphyre: These are actually exact replicas of two tests further down in the file. Pretty sure this was originally me with fat ctrl-v fingers, sorry!,

rarruda: good catch!

spec/unleash/constraint_spec.rb
sighphyre: What about `it 'gracefully handles invalid constraint operators'` instead? I read this as handling gracefully invalid operators and I'm not immediately sure how something can be gracefully invalid

rarruda: Good point! Your suggested wording is even better/clearer! Love it!

lib/unleash/constraint.rb
sighphyre: I like the first suggestion! I think we should rather unpack a bit in the comment why this behaviour is here, because I think this is a surprising edge case in constraints on the Unleash side and its helpful to understand why this code is necessary

rarruda: Agree!

spec/unleash/constraint_spec.rb
gardleopard: done

lib/unleash/constraint.rb
```diff
@@ -80,8 +92,8 @@ def self.on_valid_version(val1, val2)
 
     # This should be a private method but for some reason this fails on Ruby 2.5
     def validate_constraint_value_type(operator, value)
-      raise ArgumentError, "context_name is not an Array" if LIST_OPERATORS.include?(operator) && value.is_a?(String)
-      raise ArgumentError, "context_name is not a String" if !LIST_OPERATORS.include?(operator) && value.is_a?(Array)
+      Unleash.logger.info "value is not an Array, operator is expecting an Array" if LIST_OPERATORS.include?(operator) && value.is_a?(String)
+      Unleash.logger.info "value is an Array, operator is not an Array operator" if !LIST_OPERATORS.include?(operator) && value.is_a?(Array)
```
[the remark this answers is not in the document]
gardleopard: Im not sure it makes sense, because I considered just removing this. Reason is that this test should really live inside unleash. WDYT?

spec/unleash/constraint_spec.rb
gardleopard: I'm unsure if we should have code for this at all, as this should be handled by unleash. WDYT?

rarruda: I agree that the unleash server should never send us invalid data, but a user could still provide us invalid data via the bootstrap functionality. Additionally the SDK should attempt to be as robust as possible.

So I would really recommend to keep it.

lib/unleash/constraint.rb
```diff
@@ -80,8 +88,10 @@ def self.on_valid_version(val1, val2)
 
     # This should be a private method but for some reason this fails on Ruby 2.5
     def validate_constraint_value_type(operator, value)
-      raise ArgumentError, "context_name is not an Array" if LIST_OPERATORS.include?(operator) && value.is_a?(String)
-      raise ArgumentError, "context_name is not a String" if !LIST_OPERATORS.include?(operator) && value.is_a?(Array)
+      if LIST_OPERATORS.include?(operator) && value.is_a?(String)
+        Unleash.logger.info "value is not an Array, operator is expecting an Array"
+      end
```
rarruda, line 93:
```suggestion
      Unleash.logger.info "value is a String, operator is expecting an Array" if LIST_OPERATORS.include?(operator) && value.is_a?(String)
```

lib/unleash/constraint.rb
```diff
@@ -80,8 +88,10 @@ def self.on_valid_version(val1, val2)
...
+      Unleash.logger.info "value is an Array, operator is not an Array operator" if !LIST_OPERATORS.include?(operator) && value.is_a?(Array)
```
rarruda, line 94:
```suggestion
      Unleash.logger.info "value is an Array, operator is expecting a String" if !LIST_OPERATORS.include?(operator) && value.is_a?(Array)
```

spec/unleash/constraint_spec.rb
gardleopard: Ok will give it a stab

gardleopard: Fixed in latest commit:)

lib/unleash/constraint.rb
```diff
@@ -45,6 +49,10 @@ def matches_context?(context)
     rescue KeyError
       Unleash.logger.warn "Attemped to resolve a context key during constraint resolution: #{self.context_name} but it wasn't \
       found on the context"
+
+      # when the operator is NOT_IN and there is no data, return true. In all other cases the operator doesn't match.
+      return true if operator == :NOT_IN
```
rarruda, line 54: Think about it again, this seems to belong in matches_constraint? Method and not in the matches_context?


So for this operator the matches_constraint? Should not throw an exception. (But still should for all other operators).

Also note that this doesn't seem to respect the inverted method parameter. Noting it up the stack should solve that. But maybe add a trivial assert for NOT using the inverted parameter (in addition to the current one using)?

spec/unleash/constraint_spec.rb
rarruda: Thinking about it twice, maybe we can move this clean up snippet to a separate trivial clean up PR?

spec/unleash/constraint_spec.rb
```diff
@@ -48,6 +42,16 @@
       expect(constraint.matches_context?(context)).to be true
     end
 
+    it 'matches based on a value NOT_IN in empty context field' do
```
rarruda, line 45:
```suggestion
    it 'matches based on a value NOT_IN in a not existing context field' do
```

lib/unleash/constraint.rb
```diff
@@ -80,8 +88,8 @@ def self.on_valid_version(val1, val2)
 
     # This should be a private method but for some reason this fails on Ruby 2.5
     def validate_constraint_value_type(operator, value)
```
rarruda, line 90: We should consider renaming this method, as it no longer validates, but only checks validity. And keeping as is would be misleading. 

Can't come up with a decent name now, but can try to think about a suggestion later on. Possibly we can consider in-lining it too with a comment? 🤷‍♂️ I dunno. 

Open for naming to suggestions here.

rarruda (commented):
I think I found a corner case that needs to be looked at, plus some other general comments.

lib/unleash/constraint.rb
gardleopard: done

lib/unleash/constraint.rb
gardleopard: The latest commit should solve all of this. The reversing happens in matches_context and in matches_constraint we are now handling the nil values.

spec/unleash/constraint_spec.rb
rarruda: would it be ok to put this in a separate PR, just so we have one PR per problem / issue getting fixed?

spec/unleash/constraint_spec.rb
```diff
@@ -384,23 +388,38 @@
...
+        properties: {
+          env: 'development'
+        }
+      }
+      context = Unleash::Context.new(context_params)
+      constraint = Unleash::Constraint.new('env', 'NOT_A_VALID_OPERATOR', 'dev', inverted: true)
+      expect(constraint.matches_context?(context)).to be true
+
+      constraint = Unleash::Constraint.new('env', 'NOT_A_VALID_OPERATOR', ['dev'], inverted: true)
+      expect(constraint.matches_context?(context)).to be true
```
rarruda, line 405: Maybe also test without the inverted clause?

```suggestion
      constraint = Unleash::Constraint.new('env', 'NOT_A_VALID_OPERATOR', 'dev')
      expect(constraint.matches_context?(context)).to be false

      constraint = Unleash::Constraint.new('env', 'NOT_A_VALID_OPERATOR', ['dev'])
      expect(constraint.matches_context?(context)).to be false

      constraint = Unleash::Constraint.new('env', 'NOT_A_VALID_OPERATOR', 'dev', inverted: true)
      expect(constraint.matches_context?(context)).to be true

      constraint = Unleash::Constraint.new('env', 'NOT_A_VALID_OPERATOR', ['dev'], inverted: true)
      expect(constraint.matches_context?(context)).to be true
```

or consider having just without the inverted parameter, so the test covers exactly it's objective and nothing else that is already tested elsewhere... 

```suggestion
      constraint = Unleash::Constraint.new('env', 'NOT_A_VALID_OPERATOR', 'dev')
      expect(constraint.matches_context?(context)).to be false

      constraint = Unleash::Constraint.new('env', 'NOT_A_VALID_OPERATOR', ['dev'])
      expect(constraint.matches_context?(context)).to be false
```

Feel free to pick between these two.

lib/unleash/constraint.rb
```diff
@@ -38,8 +42,6 @@ def initialize(context_name, operator, value = [], inverted: false, case_insensi
     end
 
     def matches_context?(context)
-      Unleash.logger.debug "Unleash::Constraint matches_context? value: #{self.value} context.get_by_name(#{self.context_name})" \
-        " #{context.get_by_name(self.context_name)} "
```
rarruda, line 42: When debugging, it is nice to have this message come up. 

So I would avoid dropping it, if i could...

lib/unleash/constraint.rb
```diff
@@ -92,6 +94,12 @@ def matches_constraint?(context)
         false
       end
 
+      # when the operator is NOT_IN and there is no data, return true. In all other cases the operator doesn't match.
+      return self.operator == :NOT_IN unless context.property?(self.context_name)
+
+      Unleash.logger.debug "Unleash::Constraint matches_context? value: #{self.value} context.get_by_name(#{self.context_name})" \
+      " #{context.get_by_name(self.context_name)} "
```
rarruda, line 101: should switch order here, and rename the method name within the log message / rewrite it to the log messages current context.

```suggestion
      Unleash.logger.debug "Unleash::Constraint matches_constraint?  value: #{self.value} constraint: #{self.operator} context.get_by_name(#{self.context_name})" \
      " #{context.get_by_name(self.context_name)} "

      # when the operator is NOT_IN and there is no data, return true. In all other cases the operator doesn't match.
      return self.operator == :NOT_IN unless context.property?(self.context_name)
```

spec/unleash/constraint_spec.rb
```diff
@@ -384,23 +388,38 @@
...
-          Unleash::Constraint.new('env', operator_name, '')
-        end.to raise_error
+        expect(Unleash.logger).to receive(:warn).with("value is a String, operator is expecting an Array")
+        Unleash::Constraint.new('env', operator_name, '')
       end
 
       string_constraints = ['NUM_EQ', 'NUM_GT', 'NUM_GTE', 'NUM_LT', 'NUM_LTE',
                             'DATE_AFTER', 'DATE_BEFORE', 'SEMVER_EQ', 'SEMVER_GT', 'SEMVER_LT']
       string_constraints.each do |operator_name|
         Unleash::Constraint.new('env', operator_name, '')
```
rarruda, line 420: No need to test that the creation of the object succeeded w/o an exception anymore. It now should be a safe assumption.
```suggestion
```

spec/unleash/constraint_spec.rb
```diff
@@ -384,23 +388,38 @@
...
+
+      constraint = Unleash::Constraint.new('env', 'NOT_A_VALID_OPERATOR', ['dev'], inverted: true)
+      expect(constraint.matches_context?(context)).to be true
+    end
+
+    it 'warns about constraint construction for invalid value types for operator' do
       array_constraints = ['STR_CONTAINS', 'STR_ENDS_WITH', 'STR_STARTS_WITH', 'IN', 'NOT_IN']
 
       array_constraints.each do |operator_name|
         Unleash::Constraint.new('env', operator_name, [])
```
rarruda, line 412: No need to test that the creation of the object succeeded w/o an exception anymore. It now should be a safe assumption.

```suggestion
```

spec/unleash/context_spec.rb
```diff
@@ -134,6 +134,25 @@
     expect(context.get_by_name(:CountryCode)).to eq('UK')
   end
 
+  it "checks if property is found" do
```
rarruda, line 137:
```suggestion
  it "checks if property is included in the context" do
```

rarruda (commented):
some more feedback. It is looking promising!

lib/unleash/constraint.rb
gardleopard: It is not dropped, its moved. This message is what is raising the KeyError, so if it is up here we never get to graceful handling of this.

spec/unleash/constraint_spec.rb
gardleopard: ok.

lib/unleash/constraint.rb
gardleopard: We cant log it before we validate that it is set. This is the reason the log is behind the check if property is set.

spec/unleash/constraint_spec.rb
gardleopard: done

spec/unleash/constraint_spec.rb
gardleopard: I added this test based on review you had earlier. Its not testing the object creation, its checking that unleash logs a warn message.

spec/unleash/constraint_spec.rb
gardleopard: I added this test based on review you had earlier. Its not testing the object creation, its checking that unleash logs a warn message.

lib/unleash/constraint.rb
rarruda: why not keep it in both places? 

This is debug logging to help us get an idea of the state the SDK is in, if we ever need to debug it when it is deployed out in the wild.

It can be handy to have a confirmation in the log that we have made it so far in the code.

spec/unleash/constraint_spec.rb
rarruda: We have two `Unleash::Constraint.new` in this test (per `operator_name`). We only need to test the "sad" path, that the logger is getting called. No need to test the happy path....

That is at least what I meant to say.

spec/unleash/constraint_spec.rb
rarruda: We have two `Unleash::Constraint.new` in this test (per `operator_name`). We only need to test the "sad" path, that the logger is getting called. No need to test the happy path....

That is at least what I meant to say.

lib/unleash/constraint.rb
```diff
@@ -92,6 +94,12 @@ def matches_constraint?(context)
         false
       end
 
+      # when the operator is NOT_IN and there is no data, return true. In all other cases the operator doesn't match.
+      return self.operator == :NOT_IN unless context.property?(self.context_name)
```
rarruda:
```suggestion
      return self.operator == :NOT_IN unless context.include?(self.context_name)
```

lib/unleash/constraint.rb
rarruda: why not?  `context.get_by_name()` should return nil if no value is found.


We would like to get this debug message for all operators, including the `:NOT_IN`, and even if the value is not in the context.

Also the node regarding having `matches_constraint?` instead of `matches_context?` within the log message is still valid.

spec/unleash/constraint_spec.rb
gardleopard: ok, done.

lib/unleash/constraint.rb
gardleopard: this one is on an outdated thing, this is fixed

rarruda (approved):

sighphyre (approved):
Woot!

PR #119, opened 2022-11-03
fix: add possibility to passing user_id as integer

serhii-bodnaruk:
## About the changes
We should pass user_id as a string in Unleash::Context to make it work. In most cases user_id is an integer, so we have to convert the id into an integer. It's a little annoying to do it all the time.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -22,7 +22,7 @@ Metrics/BlockLength:
 Metrics/AbcSize:
   Max: 30
 Metrics/CyclomaticComplexity:
-  Max: 9
+  Max: 10
 Metrics/PerceivedComplexity:
   Max: 10
 
--- a/lib/unleash/context.rb
+++ b/lib/unleash/context.rb
@@ -9,7 +9,7 @@ module Unleash
 
       self.app_name    = value_for('appName', params, Unleash&.configuration&.app_name)
       self.environment = value_for('environment', params, Unleash&.configuration&.environment || 'default')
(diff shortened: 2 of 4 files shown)


sighphyre:
Hey @serhii-bodnaruk, yeah I can get behind this, this is a nice improvement. This PR has broken some of our tests though, if you're willing to look into that, I'm willing to take another look and merge it

serhii-bodnaruk:
hey @sighphyre! it's strange because all tests are green on my local (macOS). I'll try to find some ubuntu and run tests one more time.

serhii-bodnaruk:
@sighphyre maybe try to rerun CI. I'm not sure where is the problem. I've run locally both on ubuntu and macOS. All specs are green
Screenshot 2022-11-03 at 11 46 05

sighphyre:
> updt: found the problem. It was inside client-specification

Nice catch, thanks (I'm adding some notes to the README so that this doesn't bite someone else).

Rubocop is complaining though for silly reasons, mind taking the CyclomaticComplexity here up to 10 to make it happy?

serhii-bodnaruk:
> > updt: found the problem. It was inside client-specification
> 
> Nice catch, thanks (I'm adding some notes to the README so that this doesn't bite someone else).
> 
> Rubocop is complaining though for silly reasons, mind taking the CyclomaticComplexity here up to 10 to make it happy?

done. added one more test

sighphyre (approved):
LGTM

PR #121, opened 2022-11-14
increment metrics in way protected from global state

shieldo:
## About the changes

sentry

In rare cases there can be a race condition when incrementing metrics - `self.features` can be reset between lines 18 and 19 of the screenshot above if metrics are being reported up to the server concurrently, throwing a NoMethodError.

This change uses a local variable to build the new feature metric. It's possible that an increment can be missed during execution of the methods, but this is preferable to throwing an error when using a feature flag in application code.

### DIFF
--- a/lib/unleash/metrics.rb
+++ b/lib/unleash/metrics.rb
@@ -15,15 +15,19 @@ module Unleash
     def increment(feature, choice)
       raise "InvalidArgument choice must be :yes or :no" unless [:yes, :no].include? choice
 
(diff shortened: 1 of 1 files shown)


FredrikOseberg:
@shieldo Thanks for this PR! I'll bring it up with the team and get back to you.

sighphyre:
Hey @shieldo, there's a separate PR here #122. which fixes closes in a thread safe way, I'm going to close this in favour of that, if you're okay with it?

shieldo:
@sighphyre Looks good - #122 is preferable approach certainly and should deal with this issue correctly!

PR #123, opened 2022-11-15
Feat: add rbs support

rarruda:
## About the changes

Add support for type checking/hinting in the SDK.

Should be quite helpful for users if they have a powerful IDE, and for developers to ensure that we correctly handle the correct types in methods, etc.

## Discussion points

Not sure if all the essential client-code is yet in place.

Basic testing is however in place via github actions / steep


rarruda:
Note: worth while considering: https://github.com/soutaro/rbs-inline

PR #124, opened 2022-11-25
Chore: Add support for jruby 9.4

rarruda:
## About the changes

Be explicit about jruby 9.4 being supported. See https://www.jruby.org/2022/11/23/jruby-9-4-0-0

Also update a gha step, to avoid warning messages in the build.


sighphyre (approved):
Nice nice, thank you!

PR #126, opened 2022-11-28
Fix/variants dont add to toggle metrics

sighphyre:
This fixes an issue where variants don't report metrics when the toggle is disabled and makes variants increment the parent feature toggle's metrics (this is a spec change that never propagated to this SDK).

### DIFF
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -37,12 +37,11 @@ module Unleash
 
       context = ensure_valid_context(context)
 
-      return Unleash::FeatureToggle.disabled_variant unless self.enabled && am_enabled?(context)
-      return Unleash::FeatureToggle.disabled_variant if sum_variant_defs_weights <= 0
+      toggle_enabled = am_enabled?(context)
+      variant = resolve_variant(context, toggle_enabled)
 
-      variant = variant_from_override_match(context) || variant_from_weights(context, resolve_stickiness)
-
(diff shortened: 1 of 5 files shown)


lib/unleash/feature_toggle.rb
```diff
@@ -37,12 +37,11 @@ def get_variant(context, fallback_variant = Unleash::FeatureToggle.disabled_vari
 
       context = ensure_valid_context(context)
 
-      return Unleash::FeatureToggle.disabled_variant unless self.enabled && am_enabled?(context)
-      return Unleash::FeatureToggle.disabled_variant if sum_variant_defs_weights <= 0
-
-      variant = variant_from_override_match(context) || variant_from_weights(context, resolve_stickiness)
+      toggle_enabled = am_enabled?(context)
```
sighphyre, line 40: The `self.enabled` is already considered in `am_enabled?`

lib/unleash/feature_toggle.rb
```diff
@@ -37,12 +37,11 @@ def get_variant(context, fallback_variant = Unleash::FeatureToggle.disabled_vari
 
       context = ensure_valid_context(context)
 
-      return Unleash::FeatureToggle.disabled_variant unless self.enabled && am_enabled?(context)
```
sighphyre, line 40: I've shuffled this code around because if we want to be in line with the other SDKs we should still increment the variant counter for disabled

gardleopard (approved):
Great work, looks good to me:)

PR #128, opened 2022-12-07
wraps is_enabled and get_variant in a rescue clause

sighphyre:
We've had a number of cases where an exception has not been trapped and handled correctly, resulting in a crash for the system hosting the SDK. Usually these situation are due to using the SDK or server in a strange way but I still strongly feel that the SDK should never result in an exception bubbling up in an  unhandled way.

This might be a more controversial change but this wraps the `is_enabled` and `get_variant` methods in the client in a try/rescue and returns a default but disabled response in both those cases.

This is totally up for debate so comments, thoughts and angry screaming all welcome

### DIFF
--- a/spec/unleash/client_spec.rb
+++ b/spec/unleash/client_spec.rb
@@ -300,6 +300,12 @@ RSpec.describe Unleash::Client do
       unleash_client.is_enabled?('any_feature5', {}) { nil }
     ).to be false
 
+    expect(
+      unleash_client.is_enabled?('any_feature5', {}) do
+        raise StandardError, "Oops, evaluating a feature should not have done this!"
+      end
+    ).to be false
+
     # should never really send both the default value and a default block,
     # but if it is done, we OR the two values
     expect(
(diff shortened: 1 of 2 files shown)


spec/unleash/client_spec.rb
```diff
@@ -300,6 +300,12 @@
       unleash_client.is_enabled?('any_feature5', {}) { nil }
     ).to be false
 
+    expect(
```
gardleopard, line 303: Can you pull this out to a separate test and add cases where fallback blocks are sent in as well?
If fallback blocks are failing its not up to the sdk to handle that.

If there is a fallback block we should fallback to that.

gardleopard (changes requested):

spec/unleash/client_spec.rb
sighphyre: Hmmm, it's going to be hard to make this fail without using the block but yes, I do agree that the correct behaviour here is probably to fallback to the er... fallback. I'll patch it

rarruda:
Just to try to understand: the issue you are trying to handle is when a codeblock provided raises an exception? Do you have examples of the errors happening out in the wild?

Also if you are to log this, it is most definitely not a WARN, but at least an ERROR.

Hiding exceptions with global rescue for all errors is not usually seen as best practice, as it makes debugging incredibly difficult.

sighphyre:
> Just to try to understand: the issue you are trying to handle is when a codeblock provided raises an exception? Do you have examples of the errors happening out in the wild?

Gard has pointed out that the test I wrote is making the intent way less clear than it should be (I was horribly abusing the codeblock as a way of simulating an unexpected exception). 

The rescue has less to do with handling janky stuff in the codeblocks and more to do with programming errors that occasionally slip in. Ideally the `is_enabled` calls should never bubble an exception and risk crashing the host processes.

sighphyre:
> Hiding all possible exceptions with a global rescue for all errors is not usually seen as best practice, as it makes debugging incredibly difficult.

This is part of why this is a controversial change. I think this becomes significantly more debatable when dealing with unobtrusive libraries. Is it better to default a feature or crash a process in client code?

PR #129, opened 2022-12-25
chore: add ruby 3.2 support in gh actions

rarruda:
and document that it is a supported version

https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/


gastonfournier:
The PR looks good but it is failing with something that seems unrelated when building for Ruby 2.5:
```
  	The last version of bundler (~> 2.0) to support your Ruby & RubyGems was 2.3.26. Try installing it with `gem install bundler -v 2.3.26`
  	bundler requires Ruby version >= 2.6.0. The current ruby version is 2.5.0.
```

I'm not quite sure why this is happening (not a ruby expert), but I looked at previous builds and the same bundler version with the same ruby version seems to work fine: https://github.com/Unleash/unleash-client-ruby/actions/runs/3639016233/jobs/6141903265#step:3:20 

I'll happily help with this but I don't know how to solve this right now. If you can help, we can work it out or else we should wait for @sighphyre who knows Ruby better than me

gastonfournier (approved):
After 2 retries of the job, it succeeded. Doesn't make me feel confident, but 2.5 (the build failing) was unchanged by this PR so :+1:

PR #130, opened 2022-12-27
Adding disabled? method

serhii-bodnaruk:
## About the changes
Adding `disabled?` method to more comfort using Unleash.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,7 @@ AllCops:
 Naming/PredicateName:
   AllowedMethods:
     - is_enabled?
+    - is_disabled?
 
 
 Metrics/ClassLength:
--- a/README.md
+++ b/README.md
@@ -125,6 +125,12 @@ if @unleash.is_enabled?(feature_name, unleash_context)
 else
   puts " #{feature_name} is disabled according to unleash"
 end
+
+if @unleash.is_disabled?(feature_name, unleash_context)
+  puts " #{feature_name} is disabled according to unleash"
+else
+  puts " #{feature_name} is enabled according to unleash"
+end
 ```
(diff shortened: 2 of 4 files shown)


serhii-bodnaruk:
@sighphyre could you take a look?

rarruda:
@sighphyre @gardleopard 

Opinions?

The things I can think of:

- not a method in the other SDKs (?!), so it would make the rundt SDK a bit off compared with the API in the other SDKs. (Looked at node, Java, and python SDKs)
- Maybe the recommended solution instead of `if client.disabled?("foo")` would be to use the `unless` keyword instead `unless client.enabled?("foo")`  (?!)
- Missing clone method `is_disabled?`, for the sake of consistency. 
- Missing opposite test, for completeness sake. (That the disabled is the opposite in both true and false scenarios).
- Missing docs in README (method description).

Wdyt?

serhii-bodnaruk:
It's not looking good and well readable "unless" staff, we prefer not to use it in simple cases.
From the beginning I added the method `is_disabled?` but your CI and rubocop setup doesn't like it, so renamed the method and remove the alias.

I can add more tests and README, my bad. But in case you guys like the idea in general. If not we will patch it on our end, no problem. 🙏

rarruda:
Those were my 2c/thoughts.

Let's wait for what @sighphyre and @gardleopard think/decide. 

But maybe @ivarconr would want to weigh in too?

gardleopard (changes requested):
Hi, this looks good! Thank you for your contribution. We would like to see some updates in the README.md and then its good:)

gardleopard (approved):
Looks good, thanks for your contribution!

PR #133, opened 2023-01-24
Update murmurhash3 minimum version to 0.1.7

malclocke:
This ensures that murmurhash3 will work correctly with native libraries on current Rubygems versions, see
https://github.com/funny-falcon/murmurhash3-ruby/issues/10

## About the changes

The `murmurhash3` native library is broken on recent versions of Rubygems (3.4+) and causes the following warning:
```
Attention: used pure ruby version of MurmurHash3
``` 
The `murmerhash3` gem has been updated with a fix, and updating this dependency ensures the updated version will be installed.

See:
- https://github.com/funny-falcon/murmurhash3-ruby/issues/10
- https://github.com/rubygems/rubygems/issues/6205

### DIFF
--- a/unleash-client.gemspec
+++ b/unleash-client.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
   spec.required_ruby_version = ">= 2.5"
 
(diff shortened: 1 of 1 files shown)


rarruda (approved):
Nice catch! Thanks!

PR #134, opened 2023-02-21
docs: remove refs to `environment` in samples, update description

thomasheartman:
This change removes any references to setting `environment` in the SDK
config manually.

It also updates the config property description to clarify that it is
not the same as the "Unleash environment".

Why? Because this is a returning point of confusion for our users,

## Question

The config option description says "not yet in use". Is still true?


sighphyre (approved):
LGTM, pretty sure this is a hangover from the days before we encoded the environment in the API token

PR #136, opened 2023-03-22
Add Context#to_h

### LINKED ISSUE #135: Provide a #to_h method for Unleash::Context
### Describe the feature request

The instance of context should be able to be converted to a hash.
The ruby naming convention is a `#to_h` method to do so.

### Background

Sometimes, one wants to re-use the context set. e.g. to report it to Sentry or any other exception tracker.

### Solution suggestions

Add a `to_h` method to `Unleash::Context`

tmaier:
## About the changes
Add a `to_h` method to Context.

Closes #135

### DIFF
--- a/lib/unleash/context.rb
+++ b/lib/unleash/context.rb
@@ -22,6 +22,10 @@ module Unleash
       "<Context: user_id=#{@user_id},session_id=#{@session_id},remote_address=#{@remote_address},properties=#{@properties}" \
         ",app_name=#{@app_name},environment=#{@environment}>"
     end
+    
+    def to_h
+      ATTRS.map { |attr| [attr, self.send(attr)] }.to_h.merge(properties: @properties)
+    end
 
     def get_by_name(name)
       normalized_name = underscore(name).to_sym


rarruda:
Agree! This is definitely nice to have! 

Should be ok to merge once rubocop is happy. 😊

tmaier:
I think i've fixed the style errors

tmaier:
@rarruda I meeged your suggestion re. Array#to_h.

however, this is only supported for Ruby >= 2.6
Ruby 2.6 and below already hit EoL. 2.7 is about to hit EoL in April this year.
I can either revert this last commit or you could remove officially the support for EoL Ruby versions.
However, this is out of scope for this PR.
See <https://www.ruby-lang.org/en/downloads/branches/>

rarruda:
I should have tested the suggestion on 2.5 before mentioning it. Sorry about the extra work. My bad.

Let's revert and we can merge it as it was.

tmaier:
@rarruda no worries. I removed that last commit.

rarruda:
@tmaier thanks for your contribution!

PR #140, opened 2023-05-02
fix: forcefully disabling metrics when client is disabled (#139)

### LINKED ISSUE #139: When client is disabled, but metrics are not, metrics are referenced but never initialized.
### Describe the bug

When disabling the client but not disabling metrics, the metrics attribute/container is never initialized. Yet, attempting to call `enabled?` on the client results in `undefined method ... for nil:NilClass` as it attempts to access metrics.

### Steps to reproduce the bug

```ruby
Unleash::Client.new(
  disable_client: true,
  url: 'https://example.com',
  bootstrap_config: Unleash::Bootstrap::Configuration.new(
    data: '{
      "version": 1,
      "features": [
        {
          "enabled": true,
          "name": "featureX"
        }
      ]
    }'
  )
).enabled?('featureX', nil)
```

### Expected behavior

No runtime errors, and correct state of the feature.

### Logs, error output, etc.

```bash
[2023-05-01T18:45:55.551253+00:00          Unleash WARN ] : Unleash::Client is disabled! Will only return default (or bootstrapped if available) results!
/shared/unleash/lib/unleash/feature_toggle.rb:30:in `is_enabled?': undefined method `increment' for nil:NilClass (NoMethodError)

      Unleash.toggle_metrics.increment(name, choice) unless Unleash.configuration.disable_metrics
                            ^^^^^^^^^^
        from /shared/unleash/lib/unleash/client.rb:50:in `is_enabled?'
        from unleash_client.rb:77:in `<main>'
```

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

Ruby SDK 4.4.2

### Subscription type

Enterprise

### Hosting type

Hosted by Unleash

### SDK information (language and version)

Ruby SDK 4.4.2

rarruda:
## About the changes
Disabling the client is meant for when a user will be testing their code, and no connection to a remote server is desired.

It is hard to see a use case for collecting and reporting metrics if the client is disabled. Forcefully disabling metrics when disabling the client makes sense and is a more coherent behavior.

Closes #139

Note that this change does change behavior, but to make things more sane.

### Important files
* `client.rb`
* `client_spec.rb`

## Discussion points
Also added `WebMock.reset!` in the `client_spec.rb` to ensure no leakage between tests.

Maybe it would have made more sense to have it overridden in `lib/unleash/configuration.rb` (by moving `disable_metrics` from `attr_accessor` to `attr_writter` + own getter) ?

### DIFF
--- a/lib/unleash/client.rb
+++ b/lib/unleash/client.rb
@@ -21,7 +21,7 @@ module Unleash
       Unleash.toggle_fetcher = Unleash::ToggleFetcher.new
       if Unleash.configuration.disable_client
         Unleash.logger.warn "Unleash::Client is disabled! Will only return default (or bootstrapped if available) results!"
-        Unleash.logger.warn "Unleash::Client is disabled! Metrics and MetricsReporter also are disabled!"
+        Unleash.logger.warn "Unleash::Client is disabled! Metrics and MetricsReporter are also disabled!"
         Unleash.configuration.disable_metrics = true
         return
       end
--- a/lib/unleash/client.rb
+++ b/lib/unleash/client.rb
@@ -21,6 +21,8 @@ module Unleash
       Unleash.toggle_fetcher = Unleash::ToggleFetcher.new
       if Unleash.configuration.disable_client
         Unleash.logger.warn "Unleash::Client is disabled! Will only return default (or bootstrapped if available) results!"
+        Unleash.logger.warn "Unleash::Client is disabled! Metrics and MetricsReporter also are disabled!"
+        Unleash.configuration.disable_metrics = true
         return
       end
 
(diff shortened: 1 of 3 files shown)


lib/unleash/client.rb
```diff
@@ -21,6 +21,8 @@ def initialize(*opts)
       Unleash.toggle_fetcher = Unleash::ToggleFetcher.new
       if Unleash.configuration.disable_client
         Unleash.logger.warn "Unleash::Client is disabled! Will only return default (or bootstrapped if available) results!"
+        Unleash.logger.warn "Unleash::Client is disabled! Metrics and MetricsReporter also are disabled!"
```
thomasheartman, line 24:
```suggestion
        Unleash.logger.warn "Unleash::Client is disabled! Metrics and MetricsReporter are also disabled!"
```

thomasheartman (commented):
Yeah, this looks pretty good to me! 😄  Some minor adjustment suggestions that you can choose to accept or reject. 

Regarding the comment in the description:
> Note that this change does change behavior, but to make things more sane.

Would you classify it as a bugfix? If not: is it a breaking change? From what I can tell, the previous behavior didn't really work and probably wasn't intended?

sighphyre (approved):
Nice! LGTM

PR #142, opened 2023-05-13
chore: update gems: rspec, webmock and rubocop

rarruda:
## About the changes
Update development gems.

Unfortunately these changes are tightly coupled, so a smaller PR is not really feasible.

## Discussion points
Updated github actions/checkout  to a supported version.

Updated `Metrics/ClassLength` in rubocop config so PR in #138 should not fail.

Note that rubocop now will only work on supported versions of ruby. Given that this is only for when developmening/testing, it should be fair to have the `if RUBY_VERSION` around it, even if in principle that's not ideal. The other option would be to stop supporting ancient versions of ruby, but that's not an adequate trade-off. 

Linted some code so the new rubopcop is happy.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -38,6 +38,8 @@ jobs:
           - 2.6
           - 2.5
 
+    needs:
+      - lint
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


sighphyre (approved):
Rubocop being Ruby version dependent is a bit sad

PR looks good though, agree the `if` is probably less painful than it looks

PR #143, opened 2023-05-13
Fix handling of invalid context in get_variant

rarruda:
## About the changes

Should be merged ONLY after #138, #142 are merged.

Adds test cases that #138 did not cover. #138 should soon have the fix in `feature_toggle.rb` that is included here. (but I  will be remove from this PR when rebasing once #138 is merged). The fix is included here to show that the tests fail w/o it, and pass with it.

Also added a comment in `get_by_name()` in `context.rb` as a reminder that it will throw exceptions when a key is not found (IIRC by design).

PR #142 fixes the rubocop issues found here.

Kudos to @jdurkee-mdsol for finding the bug where we didn't handle the case for get_variant when the context wasn't valid.

### DIFF
--- a/lib/unleash/context.rb
+++ b/lib/unleash/context.rb
@@ -27,6 +27,7 @@ module Unleash
       ATTRS.map{ |attr| [attr, self.send(attr)] }.to_h.merge(properties: @properties)
     end
 
+    # returns the value found for the key in the context, or raises a KeyError exception if not found.
     def get_by_name(name)
       normalized_name = underscore(name).to_sym
 
--- a/spec/unleash/feature_toggle_spec.rb
+++ b/spec/unleash/feature_toggle_spec.rb
@@ -506,7 +506,7 @@ RSpec.describe Unleash::FeatureToggle do
     end
   end
 
-  describe 'FeatureToggle with custom stickiness' do
+  describe 'FeatureToggle variant with custom stickiness' do
     let(:feature_toggle) do
       Unleash::FeatureToggle.new(
         "name" => "toggleName",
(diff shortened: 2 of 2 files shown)


rarruda:
Will change from draft PR to ready for review once #138 is merged.

lib/unleash/feature_toggle.rb
```diff
@@ -97,10 +97,16 @@ def sum_variant_defs_weights
     end
 
     def variant_salt(context, stickiness = "default")
-      return context.get_by_name(stickiness) unless stickiness == "default"
-      return context.user_id unless context.user_id.to_s.empty?
-      return context.session_id unless context.session_id.to_s.empty?
-      return context.remote_address unless context.remote_address.to_s.empty?
+      begin
+        return context.get_by_name(stickiness) if !context.nil? && stickiness != "default"
```
sighphyre, line 101: So I noticed that the previous PR passes the spec tests without this. Is this sneaky undefined behaviour we don't have codified in the spec tests?

sighphyre (approved):
Nice catch, LGTM

sighphyre (approved):
Nice catch, LGTM

lib/unleash/feature_toggle.rb
rarruda: Correct!

You can see from the build logs in the commit at https://github.com/Unleash/unleash-client-ruby/commit/ff28b3100d5c3cb6611d2ea145531d386272ff95  (tests w/o fix) that it fails w/o the fix due to lack of relevant tests. The tests come in the following commit.

sighphyre: Oof. Right, that needs to make its way into the spec tests. Not relevant to this PR. 

Looks like a sane behaviour, I support this!

Issue #94, opened 2022-04-04
Support semantic versioning for releases


pravi:
Currently README recommends using ~> 4.0.0. Please follow https://semver.org and update README to explicitly state compliance with semver. This will allow a looser dependency specification ~> 4.0

rarruda:
This is in place now as described in releasing section of the README. We are also adding a CHANGELOG to help keep track of changes between releases as well.

Thanks for your feedback and help!

pravi:
@rarruda thanks!

pravi:
Though installation section still recommends,
gem 'unleash', '~> 4.4.0'

which could be just ~> 4.4

PR #144, opened 2023-05-23
chore: docs: Add Changelog and update docs sections on releasing and contributing (#79) (#94)

### LINKED ISSUE #79: Release v4.0.0. What is the braking change?
Hi,

you recently released v4.0.0.
What are the breaking changes a user needs to be aware of when upgrading?

I checked the diff, but it is not obvious.
https://github.com/Unleash/unleash-client-ruby/compare/v3.2.5...v4.0.0

Consider to use the GitHub release feature. A nice way to communicate changes is https://keepachangelog.com

Kind regards

rarruda:
## About the changes
SSIA

Closes #79 && #94

## Discussion points
1) Would like feedback from the Unleash team regarding the use of Changelog (see #79 ).

 It is only useful if it is kept up-to-date, so it does require a bit more effort in documenting changes. Not sure what is the practice in the other SDKs, or if you guys have opinions regarding this. (maybe referring to the gitlog is a good enough policy?) 

Only went back to release 4.0.0 when recreating the changelog history.

2) Also thought it would be nice to clarify the steps when creating a new release.

Thought it would be nice to have these in place before creating the next release. :slightly_smiling_face:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ Note: These changes are not considered notable:
 
 ## [Unreleased]
 
+## [4.4.3] - 2023-06-14
+
+### Fixed
+- Variant metrics are now sent correctly to Unleash. Fixed a typo in the payload name. (#145)
+
 ## [4.4.3] - 2023-05-23
 ### Added
 - Add Context#to_h method (#136)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-Note: These changes are not considered notable: 
+Note: These changes are not considered notable:
 - build
 - documentation
 - dependencies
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,97 @@
...
+- Allow context to correctly resolve camelCase property values (#74)
+- Avoid unlikely situation of config changing under the read operation
+- backup path file being incorrectly set
+
+### Changed
+- change how we handle the server api url (avoid double slashes in urls used for API calls.)
+- default values: refresh_interval => 10, metrics_interval=> 30
+- changed metrics reporting behavior
+- only send metrics if there is data to send. (#58)
+
```
rarruda, line 87: I think i missed commit 31e85f3 :see_no_evil:

thomasheartman (approved):
I think this looks great! And big kudos for documenting all changes back to v4 🏆 

Regarding your questions:
1. I think having a curated changelog is super useful. Yes, it requires some more work, but I think it's worthwhile. I don't think we have a standard for this anywhere. The closest is Unleash proper, which is moving away from a git log-generated changelog to a manually created one (similar to what you've done here 🙌🏼 ).
2. Steps on how to create a release are also very much appreciated! Super helpful!

So in other words: I'm very on board with this. It may be that the others have some thoughts on changelogs in other SDKs, though, so we **could** wait and see what they say 💁🏼

rarruda:
Let's wait for another day then.

I see that the python SDK has also a [changelog](https://github.com/Unleash/unleash-client-python/blob/main/CHANGELOG.md) but follows a different standard. Not sure if standardizing is as important as having it (or not). :shrug:

sighphyre:
Basically what Thomas says - each SDK has it's own patterns around this (honestly it's own patterns around everything) I don't think they need to remain in lockstep. But I support having a changelog, it's definitely nice for users to be able to find this information without asking so LGTM!

sighphyre (approved):
LGTM!

thomasheartman:
@rarruda Is there anything left to do before merging this? It's been siting around for a bit, so just wanted to check in ☺️

rarruda:
There are a couple of changes missing links to their PRs and at least one typo still. Also I've had very little free time lately. 

I will try to get this one out very soon so we can have the next release including pr #145 already here.

thomasheartman:
Understood! Would you be willing to merge it now and update the remaining links and typos later or would you rather wait to get it fixed? And I'd be happy to put in a description of #145 under either a new 4.4.3 header or the unreleased header. Got any preferences?

rarruda:
Feel free to add to the 4.4.3 header since that's coming right around the corner! :)

thomasheartman:
Wait, sorry, I'm afraid I've made a mess. It says 4.4.3 was released on May 23, but I don't see that release on rubygems? I added the new stuff under 4.4.4 because 4.4.3 already existed. What's the right course of action here?

thomasheartman:
Update: I've gone ahead and done what I suggested. Looks like 4.4.3 was never released, so I've added it there.

PR #146, opened 2023-07-04
fix: flexible rollout strategy without context

gardleopard:
The flexible rollout strategy should evaluate default and random stickiness even if context is not provided.

### DIFF
--- a/lib/unleash/strategy/flexible_rollout.rb
+++ b/lib/unleash/strategy/flexible_rollout.rb
@@ -43,7 +43,7 @@ module Unleash
           random
         when 'default'
           return random unless context.instance_of?(Unleash::Context)
-          
+
           context.user_id || context.session_id || random
         else
           begin
--- a/lib/unleash/strategy/flexible_rollout.rb
+++ b/lib/unleash/strategy/flexible_rollout.rb
@@ -42,9 +42,7 @@ module Unleash
         when 'random'
           random
         when 'default'
-          return random unless context.instance_of?(Unleash::Context)
-
-          context.user_id || context.session_id || random
+          context&.user_id || context&.session_id || random
         else
           begin
             context.get_by_name(stickiness)
--- a/lib/unleash/strategy/flexible_rollout.rb
+++ b/lib/unleash/strategy/flexible_rollout.rb
@@ -12,7 +12,7 @@ module Unleash
         return false unless params.is_a?(Hash)
 
         stickiness = params.fetch('stickiness', 'default')
-        return false unless context.instance_of?(Unleash::Context) || ['random', 'default'].include?(stickiness)
+        return false unless context_is_sufficient?(stickiness, context)
 
         stickiness_id = resolve_stickiness(stickiness, context)
 
@@ -32,6 +32,12 @@ module Unleash
       end
 
       private
+
+      def context_is_sufficient?(stickiness, context)
+        return true if ['random', 'default'].include?(stickiness)
+
+        context.instance_of?(Unleash::Context)
+      end
 
       def random
         Random.rand(0..100)
--- a/lib/unleash/strategy/flexible_rollout.rb
+++ b/lib/unleash/strategy/flexible_rollout.rb
@@ -13,6 +13,7 @@ module Unleash
 
         stickiness = params.fetch('stickiness', 'default')
         return false unless context.instance_of?(Unleash::Context) || ['random', 'default'].include?(stickiness)
+        
         stickiness_id = resolve_stickiness(stickiness, context)
 
         begin
--- a/lib/unleash/strategy/flexible_rollout.rb
+++ b/lib/unleash/strategy/flexible_rollout.rb
@@ -12,7 +12,7 @@ module Unleash
         return false unless params.is_a?(Hash)
 
         stickiness = params.fetch('stickiness', 'default')
-        return false unless context_is_sufficient?(stickiness, context)
+        return false if context_invalid?(stickiness, context)
 
         stickiness_id = resolve_stickiness(stickiness, context)
 
@@ -33,10 +33,10 @@ module Unleash
 
       private
 
-      def context_is_sufficient?(stickiness, context)
-        return true if ['random', 'default'].include?(stickiness)
+      def context_invalid?(stickiness, context)
+        return false if ['random', 'default'].include?(stickiness)
 
-        context.instance_of?(Unleash::Context)
+        !context.instance_of?(Unleash::Context)
       end
 
       def random
--- a/lib/unleash/strategy/flexible_rollout.rb
+++ b/lib/unleash/strategy/flexible_rollout.rb
@@ -12,7 +12,7 @@ module Unleash
         return false unless params.is_a?(Hash)
 
         stickiness = params.fetch('stickiness', 'default')
-        return false unless ( context.instance_of?(Unleash::Context) || ['random', 'default'].include?(stickiness) )
+        return false unless context.instance_of?(Unleash::Context) || ['random', 'default'].include?(stickiness)
         stickiness_id = resolve_stickiness(stickiness, context)
 
         begin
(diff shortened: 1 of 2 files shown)


lib/unleash/strategy/flexible_rollout.rb
```diff
@@ -41,6 +42,8 @@ def resolve_stickiness(stickiness, context)
         when 'random'
           random
         when 'default'
+          return random unless context.instance_of?(Unleash::Context)
+
           context.user_id || context.session_id || random
```
sighphyre, line 47: Don't think we need the line above, we should be able to do

```suggestion
   context&.user_id || context&.session_id || random
```

rarruda: Given that resolve_stickiness is a private method, I agree that it is not the correct place to check that the argument is not of the correct type. It should be a safe assumption. @sighphyre 's suggestion is a good one. Using the `&. ` (safe navigation operator) is always healthy.

lib/unleash/strategy/flexible_rollout.rb
```diff
@@ -10,9 +10,10 @@ def name
       # need: params['percentage']
       def is_enabled?(params = {}, context = nil)
         return false unless params.is_a?(Hash)
-        return false unless context.instance_of?(Unleash::Context)
 
         stickiness = params.fetch('stickiness', 'default')
+        return false unless context.instance_of?(Unleash::Context) || ['random', 'default'].include?(stickiness)
```
rarruda, line 15: This is a bit complex logic. Any chance to try to extract the check into a method, and use the method name to document what the check is trying to accomplish?

lib/unleash/strategy/flexible_rollout.rb
gardleopard: ok, fixed

lib/unleash/strategy/flexible_rollout.rb
rarruda, line 15:
```suggestion
        return false if is_context_invalid?(context, stickiness)
```

```ruby
# Context is always valid when using random or default strategies.
def is_context_invalid?(context, stickiness)
    context.instance_of?(Unleash::Context) || ['random', 'default'].include?(stickiness)
end
```

rarruda (commented):
just some 2c.

lib/unleash/strategy/flexible_rollout.rb
rarruda: Feel free to rename the method, not sure it's the most appropriate. Just a suggestion.

gardleopard: I did it slightly different in order to keep the argument order and be more explicit.

lib/unleash/strategy/flexible_rollout.rb
```diff
@@ -10,9 +10,10 @@ def name
       # need: params['percentage']
       def is_enabled?(params = {}, context = nil)
         return false unless params.is_a?(Hash)
-        return false unless context.instance_of?(Unleash::Context)
 
         stickiness = params.fetch('stickiness', 'default')
+        return false unless context_is_sufficient?(stickiness, context)
```
rarruda, line 15: Any way we can simplify further by turning the `unless` to a `if` ?

Would a name like `context_required?`/`context_needed?`/`context_valid?`/`context_invalid?` be more descriptive?

lib/unleash/strategy/flexible_rollout.rb
```diff
@@ -32,6 +33,12 @@ def is_enabled?(params = {}, context = nil)
 
       private
 
+      def context_is_sufficient?(stickiness, context)
+        return true if ['random', 'default'].include?(stickiness)
```
rarruda, line 37: Note that it will not evaluate the strategy randomly when the stickiness is set to `random`. 

You sure that's the desired behaviour?

Maybe it's worth having a rspec test just for the`'stickiness' => 'radom'` case? (where a couple of different seeds are hard coded?)

gardleopard: this will just let it continue to the resolve stickiness like it did before, this is just a safe guard. It returns to the is_enabled method, not to the caller.

lib/unleash/strategy/flexible_rollout.rb
gardleopard: ok i can change it

lib/unleash/strategy/flexible_rollout.rb
rarruda: but what happens up the stack, in the `is_enabled?` method? it returns right away, no? or is the boolean logic a bit hard to read?

gardleopard: it only returns false when its supposed to, if the context is sufficient it continues...

lib/unleash/strategy/flexible_rollout.rb
```diff
@@ -10,9 +10,10 @@ def name
       # need: params['percentage']
       def is_enabled?(params = {}, context = nil)
         return false unless params.is_a?(Hash)
-        return false unless context.instance_of?(Unleash::Context)
 
         stickiness = params.fetch('stickiness', 'default')
+        return false if context_invalid?(stickiness, context)
```
rarruda, line 15: This is very easy to read!

But, Is the intention to have `is_enabled?` always return `false` when (one of):
- context is not `Unleash::Context` (ok!)
- stickiness is `default` (ok?)
- stickiness is `random` (???)

gardleopard: if stickiness is default or random continue evaluating the rollout

if stickiness is something else, require to have unleash::context

rarruda: I understand that's the desired behavior. Any chance you can add separate unit tests for these scenarios?

gardleopard: ok

lib/unleash/strategy/flexible_rollout.rb
```diff
@@ -32,6 +33,12 @@ def is_enabled?(params = {}, context = nil)
 
       private
 
+      def context_invalid?(stickiness, context)
+        return false if ['random', 'default'].include?(stickiness)
```
rarruda, line 37:
```suggestion
        # Context is always valid (and ignored) when stickiness is default or random, as context is not really relevant then.
        return false if ['random', 'default'].include?(stickiness)
```

If you don't want to add the comment here, feel free to add the equivalent in the squashed commit message.

lib/unleash/strategy/flexible_rollout.rb
```diff
@@ -41,7 +48,7 @@ def resolve_stickiness(stickiness, context)
         when 'random'
           random
         when 'default'
-          context.user_id || context.session_id || random
+          context&.user_id || context&.session_id || random
```
rarruda, line 51: note that in the special case of `stickiness => default`, it will still crash if you send a `context` of type `String` for example. But if you are doing that, you are really asking for trouble.

spec/unleash/strategy/flexible_rollout_spec.rb
```diff
@@ -5,15 +5,48 @@
...
+
+    it 'should always be disabled when stickiness is random and rollout is set to 0' do
+      params = {
+        'groupId' => 'Demo',
+        'rollout' => 0,
+        'stickiness' => 'random'
+      }
+
+      expect(strategy.is_enabled?(params, unleash_context)).to be_falsey
+      expect(strategy.is_enabled?(params, nil)).to be_falsey
```
rarruda, line 49:
```suggestion
      expect(strategy.is_enabled?(params, "invalid_context")).to be_falsey
      expect(strategy.is_enabled?(params, nil)).to be_falsey
```

gardleopard: nice catch, i fixed it now

lib/unleash/strategy/flexible_rollout.rb
gardleopard: ill add it to the commit message, the tests shows this as well

rarruda (commented):
Thanks for the patience. Looks great now!

PR #150, opened 2023-09-12
feat: add spans using OpenTelemetry

SimenB:
## About the changes

Discussed briefly on [Slack](https://unleash-community.slack.com/archives/C03GZBSJV9A/p1693828069751429).

This does 2 things.

1. Adds a span around network requests. If not, everything instrumented by `Net::HTTP` (i.e. `connect` and `HTTP GET`) get their own separate traces.

Before:
image
image
image

After:
image
image

2. Adds Semantic Conventions Attributes for feature toggle systems when evaluating toggles: https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/feature-flags/#convention
image

### Important files

Make sure to ignore whitespace when reviewing: https://github.com/Unleash/unleash-client-ruby/pull/150/files?w=1

## Discussion points

This is not a general solution - it's specifically a solution for OTel. So I'm not sure if this is acceptable to you.

FWIW, I did pretty much the same fork for the Node client in my previous job: https://github.com/folio-as/unleash-client-node/commit/f708fcd116fd614d59d968c8b6318c94f938143d?w=1

If this PR is accepted, I'm happy to clean that one up and submit a PR.

Also of note is that I only started writing ruby about a month ago after not touching it in school 10 years ago. So newbie feedback on code style, conventions etc. very much welcome 😀

### DIFF
--- a/lib/unleash/client.rb
+++ b/lib/unleash/client.rb
@@ -181,12 +181,13 @@ module Unleash
     end
 
     def add_trace_attributes(key, variant)
+      current_span = OpenTelemetry::Trace.current_span
       # OpenTelemetry::SemanticConventions::Trace::FEATURE_FLAG_* is not in the gem yet
-      OpenTelemetry::Trace::current_span.add_attributes({
-        'feature_flag.provider_name' => 'Unleash',
-        'feature_flag.key' => key,
-        'feature_flag.variant' => variant,
-      })
+      current_span.add_attributes({
+                                    'feature_flag.provider_name' => 'Unleash',
+                                    'feature_flag.key' => key,
+                                    'feature_flag.variant' => variant
+                                  })
     end
   end
 end
--- a/lib/unleash/client.rb
+++ b/lib/unleash/client.rb
@@ -3,6 +3,7 @@ require 'unleash/toggle_fetcher'
 require 'unleash/metrics_reporter'
 require 'unleash/scheduled_executor'
 require 'unleash/feature_toggle'
+require 'unleash/util/open_telemetry'
 require 'unleash/util/http'
 require 'logger'
 require 'time'
@@ -12,23 +13,25 @@ module Unleash
     attr_accessor :fetcher_scheduled_executor, :metrics_scheduled_executor
 
     def initialize(*opts)
-      Unleash.configuration = Unleash::Configuration.new(*opts) unless opts.empty?
-      Unleash.configuration.validate!
+      Unleash::Util::Tracer.in_span('Unleash::Client#initialize') do |_span|
+        Unleash.configuration = Unleash::Configuration.new(*opts) unless opts.empty?
+        Unleash.configuration.validate!
 
-      Unleash.logger = Unleash.configuration.logger.clone
-      Unleash.logger.level = Unleash.configuration.log_level
+        Unleash.logger = Unleash.configuration.logger.clone
+        Unleash.logger.level = Unleash.configuration.log_level
 
-      Unleash.toggle_fetcher = Unleash::ToggleFetcher.new
-      if Unleash.configuration.disable_client
-        Unleash.logger.warn "Unleash::Client is disabled! Will only return default (or bootstrapped if available) results!"
-        Unleash.logger.warn "Unleash::Client is disabled! Metrics and MetricsReporter are also disabled!"
-        Unleash.configuration.disable_metrics = true
-        return
+        Unleash.toggle_fetcher = Unleash::ToggleFetcher.new
+        if Unleash.configuration.disable_client
+          Unleash.logger.warn "Unleash::Client is disabled! Will only return default (or bootstrapped if available) results!"
+          Unleash.logger.warn "Unleash::Client is disabled! Metrics and MetricsReporter are also disabled!"
+          Unleash.configuration.disable_metrics = true
+          return
+        end
+
+        register
+        start_toggle_fetcher
+        start_metrics unless Unleash.configuration.disable_metrics
       end
-
-      register
-      start_toggle_fetcher
-      start_metrics unless Unleash.configuration.disable_metrics
     end
 
     def is_enabled?(feature, context = nil, default_value_param = false, &fallback_blk)
@@ -44,12 +47,19 @@ module Unleash
 
       if toggle_as_hash.nil?
         Unleash.logger.debug "Unleash::Client.is_enabled? feature: #{feature} not found"
+
+        add_trace_attributes(feature, default_value)
+
         return default_value
       end
 
       toggle = Unleash::FeatureToggle.new(toggle_as_hash, Unleash&.segment_cache)
 
-      toggle.is_enabled?(context)
+      result = toggle.is_enabled?(context)
+
+      add_trace_attributes(feature, result)
+
+      result
     end
 
     def is_disabled?(feature, context = nil, default_value_param = true, &fallback_blk)
@@ -168,6 +178,15 @@ module Unleash
 
     def first_fetch_is_eager
       Unleash.configuration.use_bootstrap?
+    end
+
+    def add_trace_attributes(key, variant)
+      # OpenTelemetry::SemanticConventions::Trace::FEATURE_FLAG_* is not in the gem yet
+      OpenTelemetry::Trace::current_span.add_attributes({
+        'feature_flag.provider_name' => 'Unleash',
+        'feature_flag.key' => key,
+        'feature_flag.variant' => variant,
+      })
     end
   end
 end
--- a/lib/unleash/metrics_reporter.rb
+++ b/lib/unleash/metrics_reporter.rb
@@ -1,5 +1,6 @@
 require 'unleash/configuration'
 require 'unleash/metrics'
+require 'unleash/util/open_telemetry'
 require 'net/http'
 require 'json'
 require 'time'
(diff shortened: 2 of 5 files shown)


rarruda (commented):
I'm not quite sure we want to add more dependencies into this gem. 

Specially since it is required for quite a corner case situation/specific use-case.

I imagine you have considered monkey patching adding the tracing? I can provide examples sometime later. An idea would be having it as a wrapper gem... Just stating out ideas/discussing. No decisions here.

SimenB:
I wouldn't say tracing is a corner-case, but I very much understand reluctance towards coupling with OTel specifically 🙂 

I'd be happy to use a wrapper gem of some sort if you have some pointers on how to put one together. But I don't wanna maintain a published gem, so if this is not acceptable we'll probably just stick to using a fork (which would be this branch 😅)

rarruda:
I wrote a non-working prototype in a WIP branch here: https://github.com/Unleash/unleash-client-ruby/compare/feat/opentelemetry_wrapper with this single commit: https://github.com/Unleash/unleash-client-ruby/commit/440ab9bf073aff3d36d57ed4c68c96660e6c4033

I didn't get the monkeypatching working there quite yet, but can let you know once I do. It should take me a few more days before i can devote more time though.

But still, in there you can get a bit of an idea of how we can extend the gem in a way that should both work for your use-case and still keep the original gem untouched.

Another possibility is always that you just monkeypatch within your code base, but that's probably not what you are looking for, and not ideal for others either.

ivarconr:
Could we provide Otel support through impression event support, or through hooks you pass in as confuiguration?

SimenB:
Any news on this one? 🙂

sighphyre:
Huh, I think I gotta leave this one up to @rarruda. I definitely don't want more dependencies if we can avoid it if but I've heard requests for tracing before. I want to say having hooks that callers could tap into could let us do this without taking on the new dependency? Something like

`client.on_before_get_features(some_tracing_callback)`

But I haven't thought about this in detail

@SimenB

> I wouldn't say tracing is a corner-case

What is the use case here? I realise this sounds like a spectacularly stupid question and it very well might be but usually I'd want tracing to track down where a request is taking longer than it should. With an SDK, the poll should be completely disconnected from the evaluation (in this case by a thread), and the evaluation itself should be IO free, computationally bound and probably faster than the code that the trace needs to invoke anyway (again I apologise if this is a really dumb way of thinking about this but I kinda want a strong argument to do this for every SDK or no SDK)

rarruda:
I'm sorry that I have a hard time being more responsive.

But what you really want to do is to monkeypatch the unleash gem to add the tracing capabilities, as all you do is wrap around the HTTP calls with the opentelemetry, which in turn [monkeypatches our use of Net::HTTP](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/instrumentation.rb).

Adding open-telemetry for everyone as a dependency is not the correct solution for this case.

I have [tried a bit more getting a prototype to work](https://github.com/Unleash/unleash-client-ruby/compare/feat/opentelemetry_wrapper?expand=1) with a standalone gem, but though I think it is mostly working, It requires more polish than what i can give this late at night.

if you feel super lazy just adding these lines somewhere in your code should mostly work. (Where work, is it worked on my machine in IRB).

```ruby
require 'opentelemetry-api'
require 'unleash'

module UnleashExtensions
  module OpenTelemetry
    TRACER = ::OpenTelemetry.tracer_provider.tracer('Unleash-Client', Unleash::VERSION)
  end
end

module UnleashExtensions
  module OpenTelemetry
    module Client
    # class Client < Unleash::Client
      def initialize(*opts)
        UnleashExtensions::OpenTelemetry::TRACER.in_span('Unleash::Client#initialize') do |_span|
          super(*opts)
        end
      end

      def is_enabled?(*args)
        result = super(*args)

        add_trace_attributes(feature, result)
        result
      end

      def add_trace_attributes(key, variant)
        current_span = ::OpenTelemetry::Trace.current_span
        # OpenTelemetry::SemanticConventions::Trace::FEATURE_FLAG_* is not in the gem yet
        current_span.add_attributes({
          'feature_flag.provider_name' => 'Unleash',
          'feature_flag.key' => key,
          'feature_flag.variant' => variant
        })
      end
    end
  end
end

module UnleashExtensions::OpenTelemetry
  # class MetricsReporter < Unleash::MetricsReporter
  module MetricsReporter
    def post
      UnleashExtensions::OpenTelemetry::TRACER::Tracer.in_span('Unleash::MetricsReporter#post') do |_span|
        super
      end
    end

  end
end

module UnleashExtensions::OpenTelemetry
  # class ToggleFetcher < Unleash::ToggleFetcher
  module ToggleFetcher
    def fetch
      UnleashExtensions::OpenTelemetry::TRACER.in_span('Unleash::ToggleFetcher#fetch') do |_span|
        super
      end
    end

    def save!
      UnleashExtensions::OpenTelemetry::TRACER.in_span('Unleash::ToggleFetcher#save!') do |_span|
        super
      end
    end
  end
end

# MonkeyPatch here:
::Unleash::Client.prepend UnleashExtensions::OpenTelemetry::Client
::Unleash::MetricsReporter.prepend UnleashExtensions::OpenTelemetry::MetricsReporter
::Unleash::ToggleFetcher.prepend UnleashExtensions::OpenTelemetry::ToggleFetcher

```

Now calling the Unleash client should provide you all the open-telemetry goodness without you having to run on a fork.

If we should actually publish this monkeypatching business as a separate gem or not, I'm not sure. But I hope with this you have a starting point from where to go. And a reminder that the magic of ruby, is the power of meta-programming and that everything is dynamic and can be re-programmed on the fly. 😄

SimenB:
@rarruda thanks for the help here, and sorry for not coming back to you on it!

With a small fix, your code seems to work great for us 👍 

```diff
diff --git i/config/initializers/unleash.rb w/config/initializers/unleash.rb
index 95fbfca98..748197477 100644
--- i/config/initializers/unleash.rb
+++ w/config/initializers/unleash.rb
@@ -17,8 +17,8 @@ module UnleashExtensions
         end
       end
 
-      def is_enabled?(*args)
-        result = super(*args)
+      def is_enabled?(feature, *args)
+        result = super(feature, *args)
 
         add_trace_attributes(feature, result)
         result
```

I do think a general strategy for how to handle the async HTTP calls within the Unleash project across all SDKs would be nice, but I guess we can tackle the libraries one by one and see if any common patterns come out of it 🙂 

---

Thanks again for your help!

SimenB:
FWIW we also added a separate span for `is_enabled?` as the current decorating would override any other toggles checked within the same span.

```ruby
UnleashExtensions::OpenTelemetry::TRACER.in_span('Unleash::Client#is_enabled?') do |span|
```
instead of
```ruby
::OpenTelemetry::Trace.current_span
```

rarruda:
@SimenB do you think it would be helpful to have the example given provided in `examples/extending_unleash_with_opentelemetry.rb`  ? 

If so, I can try to add it there for future reference.

SimenB:
Yeah, I think so!

rarruda:
Awesome! I will try to add it later today then! :)

SimenB:
For completeness, this is what we'll be using (and that I've tested with success against a running instance of Tempo)


```ruby
require 'opentelemetry-api'
require 'unleash'

module UnleashExtensions
  module OpenTelemetry
    TRACER = ::OpenTelemetry.tracer_provider.tracer('Unleash-Client', Unleash::VERSION)
  end
end

module UnleashExtensions
  module OpenTelemetry
    module Client
      def initialize(*opts)
        UnleashExtensions::OpenTelemetry::TRACER.in_span('Unleash::Client#initialize') do |_span|
          super(*opts)
        end
      end

      def is_enabled?(feature, *args)
        UnleashExtensions::OpenTelemetry::TRACER.in_span('Unleash::Client#is_enabled?') do |span|
          result = super(feature, *args)

          # OpenTelemetry::SemanticConventions::Trace::FEATURE_FLAG_* is not in the gem yet
          span.add_attributes({
            'feature_flag.provider_name' => 'Unleash',
            'feature_flag.key' => feature,
            'feature_flag.variant' => result
          })

          result
        end
      end
    end
  end
end

module UnleashExtensions::OpenTelemetry
  module MetricsReporter
    def post
      UnleashExtensions::OpenTelemetry::TRACER::Tracer.in_span('Unleash::MetricsReporter#post') do |_span|
        super
      end
    end

  end
end

module UnleashExtensions::OpenTelemetry
  module ToggleFetcher
    def fetch
      UnleashExtensions::OpenTelemetry::TRACER.in_span('Unleash::ToggleFetcher#fetch') do |_span|
        super
      end
    end

    def save!
      UnleashExtensions::OpenTelemetry::TRACER.in_span('Unleash::ToggleFetcher#save!') do |_span|
        super
      end
    end
  end
end

# MonkeyPatch here:
::Unleash::Client.prepend UnleashExtensions::OpenTelemetry::Client
::Unleash::MetricsReporter.prepend UnleashExtensions::OpenTelemetry::MetricsReporter
::Unleash::ToggleFetcher.prepend UnleashExtensions::OpenTelemetry::ToggleFetcher
```

SimenB:
@rarruda another fix

```diff
diff --git i/config/initializers/unleash.rb w/config/initializers/unleash.rb
index 78bb3baed..6ae0b8d6d 100644
--- i/config/initializers/unleash.rb
+++ w/config/initializers/unleash.rb
@@ -37,7 +37,7 @@ end
 module UnleashExtensions::OpenTelemetry
   module MetricsReporter
     def post
-      UnleashExtensions::OpenTelemetry::TRACER::Tracer.in_span('Unleash::MetricsReporter#post') do |_span|
+      UnleashExtensions::OpenTelemetry::TRACER.in_span('Unleash::MetricsReporter#post') do |_span|
         super
       end
     end
```

I wish ruby had types 😅

PR #151, opened 2023-09-12
refactor: FeatureEvaluationResult struct as object

rarruda:
## About the changes
This is just a simple refactor PR, to make use of `Struct` more idiomatic (introducted in PR #148).

There is also one typo in a prosa fixed as well.

### DIFF
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -41,11 +41,11 @@ module Unleash
 
       evaluation_result = evaluate(context)
 
-      group_id = evaluation_result[:strategy]&.params.to_h['groupId'] || self.name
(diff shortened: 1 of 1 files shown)


lib/unleash/feature_toggle.rb
```diff
@@ -86,8 +86,8 @@ def evaluate(context)
           FeatureEvaluationResult.new(!strategy.nil?, strategy)
         end
 
-      Unleash.logger.debug "Unleash::FeatureToggle (enabled:#{self.enabled} " \
-        "and Strategies combined with contraints returned #{evaluation_result})"
+      Unleash.logger.debug "Unleash::FeatureToggle (enabled:#{self.enabled}) " \
+        "and Strategies combined with constraints returned #{evaluation_result})"
```
sighphyre, line 90: Ooo, nice catch!

PR #152, opened 2023-09-18
feat: yggdrasil engine

sighphyre:
## What's this?

This swaps out the logic for feature toggles in the SDK by using Yggdrasil. In particular, constructs that are now removed are:

- Evaluation logic for feature flags (this means strategies, constraints, enabled)
- Metrics calculations, but sending metrics are still handled by the SDK itself
- Custom strategies, with a bunch of glue code

## For consideration

- Perhaps controversial but I've also trimmed back the logic for custom strategies, a lot of this was to support the legacy way of registering custom strategies and has been marked as deprecated for some time. Yggdrasil won't let the caller get hooks into the custom strategies and so a lot of this isn't really possible anymore

- We have no JRuby implementation right now, that's possible with Yggdrassil but not trivial. JRuby seems fairly important so we need to make a plan for this

- Release strategy for this. It's a big change and we should be careful here


rarruda:
This is awesome work!

But as early feedback, these are my initial thoughts:

should we consider having:
1. one Gem, which a configurable engine parameter, where one could toggle between the pure ruby implementation and the rust implementation? (or eventually additionally for JRuby java/jar/jni-ruby  implementations).  
1. one Gem, where the implementation of the client would get dynamically detected and loaded. The [gem for murmur hash](https://github.com/funny-falcon/murmurhash3-ruby) we use is a good example of how it can be done for automatic loading, where the base functionality can be provided from C, Java or PureRuby w/o any configuration. 
3. two separate Gems, one with the pure Ruby implementation and one with the rust implementation. Both could co-exist in one repo, to help keep the public method signatures in sync. 

I think the current implementation should co-exist until there is full feature parity and we have enough testing done with the client using rust implementation of the engine.

There is also a question on how we would provide custom strategies, etc, but I think that is less relevant to bring up at the moment. 

Otherwise, I only have kudos to give! This is really awesome and impressive!

sighphyre:
So this:

> I think the current implementation should co-exist until there is full feature parity and we have enough testing done with the client using rust implementation of the engine.

Is an excellent idea and I'd like to do that. Yggdrasil is surprisingly more battle tested than it appears, since it forms the core of Edge and Edge has been feature complete for 6 months or so (minus custom strategies). The Ruby engine from Yggdrasil is a pretty thin wrapper around that but there's always scope for things to go wrong with such a large change. 

Medium to long term, the Ruby implementation won't be able to keep up with Yggdrasil unless it implements a full grammar parser (which I don't think is sane) so the pure Ruby implementation days are probably coming to an end in the next few months to years.  I'm not convinced about pulling the existing Ruby engine in the SDK code out into a separate gem because of that (and I also don't want to deal with potential issues with refactoring that out on top of this complexity). I had kinda of thought to revert all the deletes and toggle between Yggdrasil and the existing code within the SDK. Once we're happy it works we can cull out the existing implementation. Idle thoughts still

Custom strategies are coming! I have a blob of terribly ugly code on my machine that is a drop in replacement for custom strategies here but it needs work before more eyes see it.

sighphyre:
Shhh stalebot, shh.

Gemfile
```diff
@@ -1,4 +1,6 @@
 source 'https://rubygems.org'
 
+gem 'yggdrasil-engine', '~> 0.0.2'
```
sighphyre, line 3: Needs to move

lib/unleash/client.rb
```diff
@@ -11,14 +12,17 @@ module Unleash
   class Client
     attr_accessor :fetcher_scheduled_executor, :metrics_scheduled_executor
 
+    # rubocop:disable Metrics/AbcSize
```
sighphyre, line 15: Belongs in the Rubocop config but we probably can't get away from this one. It's the only warning generated by this PR but this initialize is literally already at the limit

lib/unleash/client.rb
```diff
@@ -11,14 +12,17 @@ module Unleash
...
     def initialize(*opts)
       Unleash.configuration = Unleash::Configuration.new(*opts) unless opts.empty?
       Unleash.configuration.validate!
 
       Unleash.logger = Unleash.configuration.logger.clone
       Unleash.logger.level = Unleash.configuration.log_level
+      Unleash.engine = YggdrasilEngine.new
```
sighphyre, line 22: I've chosen to allow the client to new up Ygg, other consumers do nil checks when they want to access it - that's the metrics handling code and the fetching code

lib/unleash/client.rb
```diff
@@ -117,7 +129,7 @@ def info
         'appName': Unleash.configuration.app_name,
         'instanceId': Unleash.configuration.instance_id,
         'sdkVersion': "unleash-client-ruby:" + Unleash::VERSION,
-        'strategies': Unleash.strategies.keys,
+        'strategies': nil,
```
sighphyre, line 132: What do we want to do here? Accessing strategies isn't really sane anymore but we can yield the names of the strategies themselves. Not actually sure what the purpose of exposing this is right now

lib/unleash/configuration.rb
```diff
@@ -40,9 +40,9 @@ def metrics_interval_in_millis
     def validate!
       return if self.disable_client
 
-      raise ArgumentError, "URL and app_name are required parameters." if self.app_name.nil? || self.url.nil?
+      raise ArgumentError, "app_name is a required parameter." if self.app_name.nil?
```
sighphyre, line 43: Bootstrap only should be a viable option (we can do this in the others), relaxing this constraint makes testing a ton easier but I may be able to roll this one back if we need

lib/unleash/context.rb
```diff
@@ -23,6 +23,22 @@ def to_s
         ",app_name=#{@app_name},environment=#{@environment}>"
     end
 
+    def as_json
```
sighphyre, line 26: For Yggdrasil to use so we can send the context as a JSON blob

spec/unleash/client_spec.rb
```diff
@@ -628,4 +627,52 @@
       end
     end
   end
+
+  it "should use custom strategies during evaluation" do
```
sighphyre, line 631: I've removed most of the legacy tests for strategies, most of them were testing that the base strategies were correctly resolved which no longer makes a lot of sense for us

lib/unleash/configuration.rb
rarruda: Valid point.

lib/unleash/metrics_reporter.rb
```diff
@@ -15,48 +14,35 @@ def initialize
     end
 
     def generate_report
-      now = Time.now
+      metrics = Unleash&.engine&.get_metrics()
```
daveleek, line 17: Why is this optional here? It doesn't seem optional in client.rb

chriswk (approved):
This looks reasonable to me. Can't wait to see Yggdrasil take over the rest of our SDKs as well :)

daveleek (approved):
Awesome!

lib/unleash/metrics_reporter.rb
sighphyre: Good question! This may be my Rust opinions coming through here. The Client is owns the engine and makes it available to everything else when it's started up. Everything else may or may not have access to it, so this is paranoia and memory and ownership hygiene

Gemfile
sighphyre: Moved to gemspec with the other dependencies

lib/unleash/client.rb
sighphyre: Okay, I've changed this, this now yields a hash of name: {strategy_name}, which is what the Unleash API is expecting - this is needed for registration

rarruda:
After looking extremely superficially at the PR:

- code coverage is not so important as an absolute metric with such a refactor. Totally ok to just ignore it in this case.
- As you support jruby natively through yggdrasil, i think it is then a good idea to drop the pluggable engine concept.
- i would recommend merging https://github.com/Unleash/unleash-client-ruby/commit/bb2d9223fa57b38aecf681a0050a00ce5a682a46 in a last release of the current major branch, so you can more easily collect usage statistics of the current and next major releases from the unleash cloud.
- i would consider creating a [beta release/pre-release gem](https://guides.rubygems.org/patterns/#prerelease-gems) before the major release of the new refactored gem, and collect feedback if possible from early adopters, before rolling out a major release. You never know what production will bring.
- i would suggest also adding to the `README.md`, info with regards on how to migrate to the new configuration format (assuming it has changed), updating the sample code in `examples/` and updating the `CHANGELOG.md`. 

all of the above can be done in a separate branch, before merging to master, but it is not strictly required.

an extra thought then is that it might also be a good idea to have a pointer in the README in the master branch a pointer to the pre release branch, so to make it more obvious what the roadmap is.

This is amazing work! :heart: 

I hope someone bakes you a cake :cake: and buys you a beer :beer: !

rarruda:
shutdown is meant as the clean shutdown method, and shutdown! As the unclean version of it. 

The idea of saving in the clean shutdown/way, so if it does get restarted later, the it would pick up where it left off if the server is unavailable. 

Most likely it's a corner case that few will meet.

sighphyre:
> shutdown is meant as the clean shutdown method, and shutdown! As the unclean version of it.
> 
> The idea of saving in the clean shutdown/way, so if it does get restarted later, the it would pick up where it left off if the server is unavailable.
> 
> Most likely it's a corner case that few will meet.

I'm amazed how closely you read these commits haha and it makes me think I need to write better commit messages 

We shouldn't need this because it gets saved as it gets read from the Unleash API. We should have the most recent version of the API state already written from the last call. I don't think saving again is going to help much unless the user is dynamically bootstrapping or something equivalently evil

rarruda:
> > Most likely it's a corner case that few will meet.
> 

It would happen in some race condition where the thread that would write to disk is not yet finished, but it is finished loading into memory, and you tell it to shutdown in between. Incredibly rare and weird.  Or some sort of strange dynamic bootstrapping, and the state doesn't get flushed to disk correctly. Not really a normal use case. As you mentioned only would happen in really strange scenarios. 

> I'm amazed how closely you read these commits haha and it makes me think I need to write better commit messages 

You have to forgive me, but sometimes I get bored during vacations. 😛

PR #154, opened 2023-10-13
refactor: use unleash client to run client specifications

gardleopard:
This is a prerequisite for testing the next version of the client specifications.

Currently the client specifications are running using the internals of the SDK, this makes them use the unleash_client instead.

### DIFF
--- a/spec/unleash/client_specification_spec.rb
+++ b/spec/unleash/client_specification_spec.rb
@@ -13,7 +13,6 @@ RSpec.describe Unleash::Client do
     Unleash.configuration = Unleash::Configuration.new
     Unleash.logger = Unleash.configuration.logger
     Unleash.logger.level = Unleash.configuration.log_level
-    Unleash.toggles = []
     Unleash.toggle_metrics = {}
 
     # Do not test metrics:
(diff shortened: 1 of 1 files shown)


spec/unleash/client_specification_spec.rb
```diff
@@ -25,22 +24,21 @@
...
             it "test that #{test['description']}" do
-              test_toggle = unleash_toggles.select{ |t| t.fetch('name', '') == test.fetch('toggleName') }.first
+              Unleash.toggles = unleash_toggles
+              Unleash.segment_cache = state_segments
 
-              toggle = Unleash::FeatureToggle.new(test_toggle, state_segments)
               context = Unleash::Context.new(test['context'])
 
-              toggle_result = toggle.is_enabled?(context)
+              toggle_result = unleash_client.is_enabled?(test.fetch('toggleName'),context)
```
sighphyre, line 41: Ooo! Can we do it for variants too? Line 55 below could use the same love

sighphyre (approved):
YASSSS!

I think we should slap it on variants too but this is a huge improvement. LGTM

PR #155, opened 2023-10-16
feat: dependant toggles

gardleopard:
This pr implements dependant feature toggles. The tests are specified in the client specifications https://github.com/Unleash/client-specification/pull/63

### DIFF
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -104,7 +104,7 @@ module Unleash
     end
 
     def get_parent(feature)
-      toggle_as_hash = Unleash&.toggles&.select{ |toggle| toggle['name'] == feature }&.first
+      toggle_as_hash = Unleash&.toggles&.find{ |toggle| toggle['name'] == feature }
       if toggle_as_hash.nil?
         Unleash.logger.debug "Unleash::Client.is_enabled? feature: #{feature} not found"
         return nil
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -77,14 +77,10 @@ module Unleash
     end
 
     def parent_dependencies_satisfied?(context)
-      return true if dependencies.empty?
-
-      dependencies.all? do |parent|
-        evaluate_parenthood(parent, context)
-      end
+      dependencies.empty? || dependencies.all?{ |parent| evaluate_parent(parent, context) }
     end
 
-    def evaluate_parenthood(parent, context)
+    def evaluate_parent(parent, context)
       parent_toggle = get_parent(parent["feature"])
 
       return false if parent_toggle.nil?
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -92,15 +92,14 @@ module Unleash
       return false unless parent_toggle.dependencies.empty?
 
       evaluation_result = parent_toggle.is_enabled?(context)
-      if parent["enabled"] == false
-        return !evaluation_result
-      else
-        return false unless evaluation_result
-      end
+      return !evaluation_result if parent["enabled"] == false
+
+      return false unless evaluation_result
 
       unless parent["variants"].nil? || parent["variants"].empty?
         return parent["variants"].include?(parent_toggle.get_variant(context).name)
       end
+
       evaluation_result
     end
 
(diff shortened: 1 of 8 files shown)


lib/unleash/feature_toggle.rb
```diff
@@ -75,9 +76,47 @@ def am_enabled?(context)
...
+      end
+
+      unless parent["variants"].nil? || parent["variants"].empty?
+        return parent["variants"].include?(parent_toggle.get_variant(context).name)
+      end
+      return evaluation_result
+    end
+
+    def get_parent(feature)
+      toggle_as_hash = Unleash&.toggles&.select{ |toggle| toggle['name'] == feature }&.first
```
kwasniew, line 106: is there Array.find method that performs select and first?

lib/unleash/feature_toggle.rb
```diff
@@ -75,9 +76,47 @@ def am_enabled?(context)
       evaluate(context).enabled?
     end
 
+    def parent_dependencies_satisfied?(context)
```
kwasniew, line 79:
```ruby
def parent_dependencies_satisfied?(context)
  dependencies.empty? || dependencies.all? { |parent| evaluate_parent(parent, context) }
end
```

lib/unleash/feature_toggle.rb
```diff
@@ -75,9 +76,47 @@ def am_enabled?(context)
...
+      return true if dependencies.empty?
+      return dependencies.all? do |parent|
+        evaluate_parenthood(parent, context)
+      end
+    end
+
+    def evaluate_parenthood(parent, context)
```
kwasniew, line 86:
```ruby
def evaluate_parent(parent, context)
  parent_toggle = get_parent(parent["feature"])
  return false if parent_toggle.nil? || !parent_toggle.dependencies.empty?

  evaluation_result = parent_toggle.is_enabled?(context)
  if parent["enabled"] == false
    return !evaluation_result
  elsif !evaluation_result
    return false
  end

  return evaluation_result if parent["variants"].nil? || parent["variants"].empty?

  parent["variants"].include?(parent_toggle.get_variant(context).name)
end
```

kwasniew (commented):
Added some minor comments

lib/unleash/feature_toggle.rb
gardleopard: yes

gardleopard: fixed

lib/unleash/feature_toggle.rb
```diff
@@ -75,9 +76,48 @@ def am_enabled?(context)
...
+      parent_toggle = get_parent(parent["feature"])
+
+      return false if parent_toggle.nil?
+
+      return false unless parent_toggle.dependencies.empty?
+
+      evaluation_result = parent_toggle.is_enabled?(context)
+      return !evaluation_result if parent["enabled"] == false
```
rarruda, line 95: nit:
```suggestion
      return !evaluation_result unless parent["enabled"]
```

lib/unleash/feature_toggle.rb
gardleopard: Thanks, fixed

lib/unleash/feature_toggle.rb
rarruda, line 91:
```suggestion
      return !parent_toggle.is_enabled?(context) unless parent["enabled"]
```

gardleopard: evaluation_result is used throughout the function several times and parent.enabled cant be used like this in the code.

kwasniew (approved):

sighphyre (changes requested):
What happens to metrics? Parent features shouldn't get metrics incremented. Can we add a test for this?

sighphyre (approved):
G. I realise we want tests for metrics, this is quite tough to do with current structure. Yggdrasil already has those covered. If we want to move to that faster in lieu of tests here that's.... probably okay

gardleopard:
> What happens to metrics? Parent features shouldn't get metrics incremented. Can we add a test for this?

We are not touching the path of metrics here. Metrics behaviour is not tested in the sdk at all. If we want to have test for metrics behaviour I vote for having that as a separate PR.

kwasniew:
@gardleopard the important part is that introducing dependent features shouldn't trigger metrics reported on the parent feature. If that's the case then we're fine. Tests can be added separately.

PR #157, opened 2023-10-18
Update Constraint#matches_constraint?

### LINKED ISSUE #156: context_value in Constraint#matches_constraint? throws FrozenError when fronzen-string-literal magic comment is true in the calling context
### Describe the bug

When frozen-string-literal magic comment is used in a file that calls the Unleash#is_enabled? method, FrozenError exception is raised if the constraint has case-insensitive enabled.

### Steps to reproduce the bug

_No response_

### Expected behavior

_No response_

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

v5.3.0

### Subscription type

Open source

### Hosting type

Self-hosted

### SDK information (language and version)

Ruby v4.5.0

thorion3006:
## About the changes

https://github.com/Unleash/unleash-client-ruby/blob/e935495d34c2647ba7af39c664930191807af034/lib/unleash/constraint.rb#L110 is a frozen string when `# frozen-string-literal: true` is used, so calling `#upcase!` throws `FrozenError`

Closes #156

### DIFF
--- a/lib/unleash/constraint.rb
+++ b/lib/unleash/constraint.rb
@@ -107,7 +107,7 @@ module Unleash
       context_value = context.get_by_name(self.context_name)
 
       v.map!(&:upcase) if self.case_insensitive
-      context_value.upcase! if self.case_insensitive
+      context_value = context_value.upcase if self.case_insensitive
 
       OPERATORS[self.operator].call(context_value, v)
     end


gardleopard (approved):
Thank you for the contribution!

lib/unleash/constraint.rb
```diff
@@ -107,7 +107,7 @@ def matches_constraint?(context)
       context_value = context.get_by_name(self.context_name)
 
       v.map!(&:upcase) if self.case_insensitive
-      context_value.upcase! if self.case_insensitive
+      context_value = context_value.upcase if self.case_insensitive
 
       OPERATORS[self.operator].call(context_value, v)
```
rarruda, line 112: would have been a prettier solution:

```suggestion
OPERATORS[self.operator].call(context_value.then{|c| self.case_insensitive ? c.upcase : c} , v.then{|c| self.case_insensitive ? c.upcase : c})
```

rarruda: note that using `!` methods in general is to be avoided, and specially on user provided data is dangerous, as you shouldn't be altering data provided / referenced by the user. (if he is to use again later on, it will be changed by this method in the SDK).

sighphyre: > the same issue will reappear if the values in v are frozen.
Unless you're doing something super weird, the values in v should only come from internal code. If that starts to fail then we have bigger problems

PR #161, opened 2023-11-06
docs/example: add example on how to add opentelemetry headers/support

rarruda:
this must be added before starting the unleash client. (in a rails initializer for example).

This is added as future reference/documentation for the solution to PR https://github.com/Unleash/unleash-client-ruby/pull/150

Few will use opentelemetry, but it is nice to have to something to point to for those that use it.


SimenB (approved):
nice 👍 

Maybe add a note to the readme as well?

PR #162, opened 2023-11-29
fix: make client resilient to non-conforming responses from `unleash-edge`

SimenB:
## About the changes

Works around https://github.com/Unleash/unleash-edge/issues/353

```
irb(main):005:0> UNLEASH.is_enabled?("some-existent-toggle")
Traceback (most recent call last):
        1: from (irb):5
NoMethodError (private method `select' called for nil:NilClass)

Stacktrace: /usr/local/lib/ruby/gems/2.7.0/gems/unleash-5.0.0/lib/unleash/feature_toggle.rb:206:in `resolve_variants'
/usr/local/lib/ruby/gems/2.7.0/gems/unleash-5.0.0/lib/unleash/feature_toggle.rb:199:in `block in initialize_strategies'
/usr/local/lib/ruby/gems/2.7.0/gems/unleash-5.0.0/lib/unleash/feature_toggle.rb:194:in `map'
/usr/local/lib/ruby/gems/2.7.0/gems/unleash-5.0.0/lib/unleash/feature_toggle.rb:194:in `initialize_strategies'
/usr/local/lib/ruby/gems/2.7.0/gems/unleash-5.0.0/lib/unleash/feature_toggle.rb:21:in `initialize'
/usr/local/lib/ruby/gems/2.7.0/gems/unleash-5.0.0/lib/unleash/client.rb:49:in `new'
/usr/local/lib/ruby/gems/2.7.0/gems/unleash-5.0.0/lib/unleash/client.rb:49:in `is_enabled?'
```

### Important files

Only one file changed

## Discussion points

Arguably this is working around a bug in `unleash-edge`, but the same `.fetch('thing', default) || []` fallback is used throughout the code base.

### DIFF
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -189,7 +189,7 @@ module Unleash
     end
 
     def initialize_strategies(params, segment_map)
-      params.fetch('strategies', [])
+      (params.fetch('strategies', []) || [])
         .select{ |s| s.has_key?('name') && Unleash.strategies.includes?(s['name']) }
         .map do |s|
           ActivationStrategy.new(
@@ -202,7 +202,7 @@ module Unleash
     end
 
     def resolve_variants(strategy)
-      strategy.fetch("variants", [])
+      (strategy.fetch("variants", []) || [])
         .select{ |variant| variant.is_a?(Hash) && variant.has_key?("name") }
         .map do |variant|
           VariantDefinition.new(


rarruda (commented):
Any chance you could add a unit test where this fails in the current code?

I'd do it myself, but probably won't be able to in the next couple of days.

spec/unleash/client_spec.rb
```diff
@@ -297,6 +297,63 @@
...
+             }
+           )
+           .to_return(status: 200, body: "", headers: {})
+
+    features_response_body = '{
+      "version": 1,
+      "features": [{
+        "name": "toggleName",
+        "enabled": true,
+        "strategies": [{ "name": "default", "constraints": [], "parameters": {}, "variants": null }],
```
SimenB, line 318: @rarruda `"variants": null` here is the thing that triggers the error

lib/unleash/feature_toggle.rb
```diff
@@ -189,7 +189,7 @@ def ensure_valid_context(context)
     end
 
     def initialize_strategies(params, segment_map)
-      params.fetch('strategies', [])
+      (params.fetch('strategies', []) || [])
         .select{ |s| s.has_key?('name') && Unleash.strategies.includes?(s['name']) }
         .map do |s|
```
rarruda, line 194: Personally, this is how i would have solved this same issue, as i find that more legible. But your approach is correct as well (and what was implemented in the other methods in the SDK below).

```suggestion
      params.fetch('strategies', [])
        &.select{ |s| s.has_key?('name') && Unleash.strategies.includes?(s['name']) }
        &.map do |s|
```

So you can implement using the safe navigation operator for bonus points, but otherwise it looks great!

(I will try to get a PR going converting the other methods to also use the safe navigation operator, for increased legibility.

rarruda (commented):
This is a great contribuition! The SDK really should takle anything thrown at it, w/o throwing exceptions. I left a note regarding the style, but I am willing to take your solution as is too if you prefer that (just note that it does risk that I might re-write later on using the safe navigation operator later on :stuck_out_tongue: ).

lib/unleash/feature_toggle.rb
rarruda: works for me!

rarruda (approved):

rarruda:
@SimenB Thanks for your awesome contribution and  making unleash better for everyone! :tada:

PR #163, opened 2024-01-02
chore: add ruby 3.3 support in gh actions

rarruda:
and document that it is a supported version

mostly because we got it as a [Christmas present from Matz](https://www.ruby-lang.org/en/news/2023/12/25/ruby-3-3-0-released/).


thomasheartman (approved):
Looks good to me! 😄

PR #168, opened 2024-03-12
feat: make user-agent headers more descriptive

rarruda:
## About the changes
include in user-agent:
- sdk version
- ruby version
- platform it is running on

did not use RUBY_DESCRIPTION as it is a bit too verbose.

## Discussion points
I would recommend getting this change in place, merged and released before #152 is merged. This should allow some sort of telemetry being in place.

I don't believe it is in the user's interest to hide this information from the server. Eventually, maybe, one could consider adding this to the registration phase in the payload as well.

### DIFF
--- a/lib/unleash/configuration.rb
+++ b/lib/unleash/configuration.rb
@@ -51,6 +51,7 @@ module Unleash
 
     def http_headers
       {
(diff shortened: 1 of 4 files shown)


sighphyre (approved):
I think long term we're going to look into making the actual metrics format more extensive so we can gather this data in a consistent way across SDKs, but this is still a great change and looks good to me

PR #169, opened 2024-03-12
chore: update github action

rarruda:
## About the changes
so it uses Node.js 20, as version 16 is getting deprecated.

## Discussion points
coverallsapp also needs to be updated, but it doesn't run in v20 upstream, so not much to do there.


sighphyre (approved):
Nice, thanks for doing this!

PR #170, opened 2024-03-12
feat: allow payload compression in POST requests

rarruda:
## About the changes
- only done if custom_http_headers config includes {'Content-Encoding' => 'gzip'}
- only allow gzip.
- 'Content-Encoding' header will get dropped in GET requests.
- undocumented feature.

## Discussion points

- could have used helper class for compressing/decompression in the style of [ActiveSuppport::Gzip](https://github.com/rails/rails/blob/6f0d1ad14b92b9f5906e44740fce8b4f1c7075dc/activesupport/lib/active_support/gzip.rb) 
- my guess is that for the vast majority of users this will not be worth enabling. But maybe for a few heavy users it could be useful. (therefore not documenting at the moment).
- not tested against a raw unleash-server. Does it support naively decompressing gzipped content?

Creating as a draft PR to gather feedback.

### DIFF
--- a/lib/unleash/util/http.rb
+++ b/lib/unleash/util/http.rb
@@ -1,5 +1,7 @@
 require 'net/http'
 require 'uri'
+require 'zlib'
+require 'stringio'
 
 module Unleash
   module Util
@@ -8,6 +10,7 @@ module Unleash
         http = http_connection(uri)
 
         request = Net::HTTP::Get.new(uri.request_uri, http_headers(etag, headers_override))
+        request.delete('Content-Encoding')
 
         http.request(request)
       end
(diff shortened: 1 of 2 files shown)


sighphyre (commented):
Meant to respond last night but I needed to check some stuff!

So we do do support Gzip compression on the metrics endpoints (not formally tested so we'll need to poke the Unleash server to confirm). So this is a nice change

Some initial thoughts - this trades CPU for data transfer. No idea how good Ruby's Gzip compression is (imagine that's C behind the scenes?) but will this be worth it for very small payloads?

The mucking about with mutable state seems unfortunate (add/remove headers dynamically). Can we do this in a cleaner way? Something like provide a `use_compression` property on the config?

PR #173, opened 2024-06-19
fix: Unexpected TypeError on constraint

marocchino:
## About the changes

The current implementation causes a type error if there is a date time in the current time. According to the [spec document](https://docs.getunleash.io/reference/unleash-context#structure), both should be possible, so we modified the implementation to allow the date time to also be passed.

Closes #

### DIFF
--- a/lib/unleash/constraint.rb
+++ b/lib/unleash/constraint.rb
@@ -57,7 +57,7 @@ module Unleash
       val1 = DateTime.parse(val1)
       val2 = val2.is_a?(DateTime) ? val2 : DateTime.parse(val2)
       yield(val1, val2)
-    rescue ArgumentError, TypeError, Date::Error
+    rescue ArgumentError, TypeError
       Unleash.logger.warn "Unleash::ConstraintMatcher unable to parse either context_value (#{val1}) \
       or constraint_value (#{val2}) into a date. Returning false!"
       false
--- a/lib/unleash/constraint.rb
+++ b/lib/unleash/constraint.rb
@@ -55,9 +55,9 @@ module Unleash
 
     def self.on_valid_date(val1, val2)
       val1 = DateTime.parse(val1)
-      val2 = DateTime.parse(val2)
+      val2 = val2.is_a?(DateTime) ? val2 : DateTime.parse(val2)
       yield(val1, val2)
-    rescue ArgumentError
+    rescue ArgumentError, TypeError, Date::Error
       Unleash.logger.warn "Unleash::ConstraintMatcher unable to parse either context_value (#{val1}) \
       or constraint_value (#{val2}) into a date. Returning false!"
       false
--- a/spec/unleash/constraint_spec.rb
+++ b/spec/unleash/constraint_spec.rb
@@ -334,6 +334,7 @@ RSpec.describe Unleash::Constraint do
       constraint = Unleash::Constraint.new('currentTime', 'DATE_AFTER', '2022-01-31T13:00:00.000Z')
       expect(constraint.matches_context?(context)).to be false
 
+    it 'matches based on property DATE_AFTER value using DateTime object in current_time' do
       context_params = {
         user_id: '123',
         session_id: 'verylongsesssionid',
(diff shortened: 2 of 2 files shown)


lib/unleash/constraint.rb
```diff
@@ -55,9 +55,9 @@ def matches_context?(context)
 
     def self.on_valid_date(val1, val2)
       val1 = DateTime.parse(val1)
-      val2 = DateTime.parse(val2)
+      val2 = val2.is_a?(DateTime) ? val2 : DateTime.parse(val2)
       yield(val1, val2)
-    rescue ArgumentError
+    rescue ArgumentError, TypeError, Date::Error
```
marocchino, line 60: According to the [spec](https://github.com/ruby/ruby/blob/master/spec/ruby/library/datetime/parse_spec.rb), it appears that only ArgumentError is returned, but in reality, a different error occurs.

```rb
$ irb
irb(main):001> require 'date_core'
=> true
irb(main):002> DateTime.parse("1")
(irb):2:in `parse': invalid date (Date::Error)

DateTime.parse("1")
               ^^^
	from (irb):2:in `<main>'
	from /Users/marocchino/.local/share/rtx/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/irb-1.13.2/exe/irb:9:in `<top (required)>'
	from /Users/marocchino/.local/share/rtx/installs/ruby/3.2.2/bin/irb:25:in `load'
	from /Users/marocchino/.local/share/rtx/installs/ruby/3.2.2/bin/irb:25:in `<main>'
irb(main):003> DateTime.parse(1)
(irb):3:in `parse': no implicit conversion of Integer into String (TypeError)

DateTime.parse(1)
               ^
	from (irb):3:in `<main>'
	from /Users/marocchino/.local/share/rtx/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/irb-1.13.2/exe/irb:9:in `<top (required)>'
	from /Users/marocchino/.local/share/rtx/installs/ruby/3.2.2/bin/irb:25:in `load'
	from /Users/marocchino/.local/share/rtx/installs/ruby/3.2.2/bin/irb:25:in `<main>'
```

rarruda: That is a very good and valid point!

rarruda, line 60: Note that `Date:Error` inherits from ArgumentError, so it should be dropped as it is rescued already:

```suggestion
    rescue ArgumentError, TypeError
```

rarruda (commented):
Any chances you could implement the suggestion to make rubocop happy?

You get bonus points if separate the test into it's own test too. :heart:

spec/unleash/constraint_spec.rb
```diff
@@ -333,6 +333,18 @@
 
       constraint = Unleash::Constraint.new('currentTime', 'DATE_AFTER', '2022-01-31T13:00:00.000Z')
       expect(constraint.matches_context?(context)).to be false
+
```
rarruda, line 336:
```suggestion
    end

```

rarruda (approved):
It was kind of you to allow me to make changes to your PR. :rocket: 

Everything looks great now! :slightly_smiling_face:

lib/unleash/constraint.rb
```diff
@@ -55,9 +55,9 @@ def matches_context?(context)
```
rarruda, line 55: Extra bike-shedding, and sort of a note to self:

Note that maybe `val1` and `val2` are inverted in relation to who is `context_value` and `constraint_value` in the logging/fallback scenario.

 Maybe we should rename `val1` and `val2` to `context_value` and `constraint_value`?

Here is a suggestion for type annotation, as the function parameters are no longer equivalent. Though maybe that would a good idea? 
```suggestion

    # @param val1 [String]
    # @param val2 [DateTime, String]
```

sighphyre (approved):
Okay I thought about it, I think this is okay. I'm fairly sure the docs that are referenced here are wrong but I think this is a sane enough change

lib/unleash/constraint.rb
sighphyre: Yeah, completely agree but I think this belongs in a separate PR

marocchino:
Thanks for the quick merge. Sorry, since CI is a confirmed and operational specification, Rubocop did not care. I will pay more attention next time.

rarruda:
@marocchino No problem at all!  

Thanks for your contribution :heart: and for allowing me to contribute to your branch!

PR #174, opened 2024-06-20
chore: no longer support MRI ruby 2.5 and jruby 9.2

rarruda:
## About the changes
As these have been EOL for many years now, and they no longer even build in modern macos. You really shouldn't be using this version of ruby at this day and age.

Expect support for MRI 2.7 and 3.0 to also be dropped in the future, as they are also EOL.

## Discussion points
I think it is ok to wait a bit more until depcecating MRI 2.6, 2.7 and 3.0... and Jruby 9.3 (which targets to be compatible with  MRI 2.6).

This fixes the current builds breaking due to the 2.5/9.2 version of ruby not even building in the latest macos builders, and feels like the cleanest solution to get the builds going again.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -30,14 +30,12 @@ jobs:
         ruby-version:
           - jruby-9.4
           - jruby-9.3
-          - jruby-9.2
           - 3.3
           - 3.2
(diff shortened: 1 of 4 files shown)


sighphyre (approved):
YESSSS

PR #175, opened 2024-06-20
Unleash::Context#to_s displays all properties

### LINKED ISSUE #172: Unleash::Context#to_s does not display all properties
I want to contribute to unleash-client-ruby, but I am not confident in my English skills, so I use ChatGPT for translation. I apologize for any strange wording or expressions :)

---

## Description

The Unleash::Context#to_s method does not display all the properties of the context object, which can lead to confusion when debugging. Specifically, the current_time property is missing in the output.

## Steps to Reproduce

Create a new Unleash::Context object in IRB:

```ruby
irb(main):001> Unleash::Context.new
```

The output will be:

```ruby
=> 
#<Unleash::Context:0x0000000107b11568
 @app_name=nil,
 @current_time="2024-06-19T15:56:37Z",
 @environment="default",
 @properties={},
 @remote_address=nil,
 @session_id=nil,
 @user_id=nil>
```

Convert the context object to a string:

```ruby
irb(main):002> Unleash::Context.new.to_s
```

The output will be:

```ruby
=> "<Context: user_id=,session_id=,remote_address=,properties={},app_name=,environment=default>"
```

## Expected Behavior

The to_s method should display all properties of the Unleash::Context object, including current_time. For example, the output should be something like:

```ruby
"<Context: user_id=, session_id=, remote_address=, properties={}, app_name=, environment=default, current_time=2024-06-19T15:56:37Z>"
```

## Actual Behavior

The to_s method currently omits the current_time property:

```ruby
"<Context: user_id=,session_id=,remote_address=,properties={},app_name=,environment=default>"
```

# Environment
- Unleash Client Version: "5.0.1"
- Ruby Version: `ruby 3.2.0 (2022-12-25 revision a528908271) [arm64-darwin21]`
- Operating System: macOS Sonoma 14.5

# Additional Information

Including all properties in the to_s method output would greatly improve the debugging experience by providing complete context information.

---

If this idea is accepted, can I create a pull request myself?

nissyi-gh:
## About the changes

- Refactored the `to_s` method in the `Context` class to dynamically include all instance variables.
- This change ensures that any new instance variables added in the future will automatically be included in the string representation without needing to modify the `to_s` method.
- This makes the method more flexible and maintainable.

RSpec and RuboCop were executed, and no issues were found related to this change.

Closes #172

### Important files

- `lib/unleash/context.rb`

## Discussion points

- None at this time.

### DIFF
--- a/lib/unleash/context.rb
+++ b/lib/unleash/context.rb
@@ -19,8 +19,11 @@ module Unleash
     end
 
     def to_s
-      "<Context: user_id=#{@user_id},session_id=#{@session_id},remote_address=#{@remote_address},properties=#{@properties}" \
-        ",app_name=#{@app_name},environment=#{@environment}>"
+      formatted_attributes = instance_variables.map do |var|
+        "#{var.to_s.delete('@')}=#{instance_variable_get(var)}"
+      end
+
+      "<Context: #{formatted_attributes.join(', ')}>"
     end
 
     def to_h
--- a/lib/unleash/context.rb
+++ b/lib/unleash/context.rb
@@ -19,11 +19,8 @@ module Unleash
     end
 
     def to_s
-      formatted_attributes = instance_variables.map do |var|
-        "#{var.to_s.delete('@')}=#{instance_variable_get(var)}"
-      end
-
-      "<Context: #{formatted_attributes.join(', ')}>"
+      "<Context: user_id=#{@user_id},session_id=#{@session_id},remote_address=#{@remote_address},properties=#{@properties}" \
+        ",app_name=#{@app_name},environment=#{@environment},current_time=#{@current_time}>"
     end
 
     def to_h


lib/unleash/context.rb
```diff
@@ -19,8 +19,11 @@ def initialize(params = {})
     end
 
     def to_s
-      "<Context: user_id=#{@user_id},session_id=#{@session_id},remote_address=#{@remote_address},properties=#{@properties}" \
-        ",app_name=#{@app_name},environment=#{@environment}>"
+      formatted_attributes = instance_variables.map do |var|
+        "#{var.to_s.delete('@')}=#{instance_variable_get(var)}"
+      end
+
+      "<Context: #{formatted_attributes.join(', ')}>"
```
rarruda, line 26: If one were to use meta-programming, i'd have preferred if the built in `to_h` method was used instead:

```suggestion
      "<Context: " + self.to_h.map{|k,v| "#{k}=#{v}"}.join(', ') + ">"
```

rarruda, line 26: Here is the simplest possible example:
```suggestion
      "<Context: user_id=#{@user_id},session_id=#{@session_id},remote_address=#{@remote_address},properties=#{@properties}" \
        ",app_name=#{@app_name},environment=#{@environment},current_time=#{@current_time}>"
```

rarruda (commented):
I would have preferred a simpler approach, without meta-programming.

Also, as in all other classes in the entire project, we do not use meta programming when writing the `to_s` method, and I would like to stick to that standard, instead of having different implementations in different classes.

I would however be positive to having meta programming in a test to verify that all instance variables are present in the `to_s` method, if you feel like doing that.

Or if using meta programming, leverage the built-in `to_h` built in method (as in the suggestion below).

nissyi-gh:
@rarruda Thank you for your review!

I learned that I should refer to implementations in other classes. I also felt that a simpler implementation without meta-programming would be better.

lib/unleash/context.rb
nissyi-gh: I had no idea to use the `to_h` method. It is a very nice and simple implementation!

nissyi-gh: Thank you very much! Simple and easy to understand. I will incorporate this suggestion!

nissyi-gh: Commit: https://github.com/Unleash/unleash-client-ruby/pull/175/commits/0734a991dbe1d1ba68a43aeb982d2170962a7aa3

nissyi-gh:
I have confirmed once again that the RSpec and RuboCop checks pass.

rarruda (approved):
Thanks for your contribution and for making the unleash gem better!

PR #176, opened 2024-07-04
chore: centralize client spec info

sighphyre:
I'd be pretty cool to have the required semver for the client-specification as a single source of truth. This PR makes that happen by encoding it in the gemspec file.

This is part of the move towards a more standardized set of metrics across all of our SDKs - I could just hard code it but that information will drift and it actually _needs_ to be accurate for Unleash to function properly.

As a part of this I've made the client-specification necessary for a successful test run. Currently leaving this in draft while I test the build but I'll put a nicer set of instructions in place before I move to ready for review


spec/unleash/client_specification_spec.rb
```diff
@@ -19,51 +19,50 @@
...
-      describe "for #{test_file}" do
-        current_test_set = JSON.parse(File.read(SPECIFICATION_PATH + '/' + test_file))
-        context "with #{current_test_set.fetch('name')} " do
-          tests = current_test_set.fetch('tests', [])
-          state = current_test_set.fetch('state', {})
-          state_features = state.fetch('features', [])
-          state_segments = state.fetch('segments', []).map{ |segment| [segment["id"], segment] }.to_h
-          let(:unleash_toggles) { state_features }
-          unleash_client = Unleash::Client.new
+  unless File.exist?(SPECIFICATION_PATH + '/index.json')
```
sighphyre, line 22: This looks like a big change, it's 99% whitespace. All this does is make the tests fail instead of warn if the client spec isn't present

gardleopard (approved):
LGTM

PR #179, opened 2024-07-08
feat: automatically generated instance_id

nunogois:
https://linear.app/unleash/issue/2-2392/make-instance-id-autogenned-ruby

This makes `instance_id` an automatically generated read-only property, instead of being set by the user.

Similar to https://github.com/Unleash/unleash-client-dotnet/pull/226

### DIFF
--- a/spec/unleash/scheduled_executor_spec.rb
+++ b/spec/unleash/scheduled_executor_spec.rb
@@ -35,18 +35,18 @@ RSpec.describe Unleash::ScheduledExecutor do
     max_exceptions = 1
 
     scheduled_executor = Unleash::ScheduledExecutor.new('TesterLoop', 0, max_exceptions)
-    new_instance_id = SecureRandom.uuid
-    original_instance_id = Unleash.configuration.instance_id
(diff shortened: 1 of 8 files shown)


spec/unleash/scheduled_executor_spec.rb
```diff
@@ -55,39 +55,39 @@
...
     end
 
     sleep 0.01
 
-    expect(Unleash.configuration.instance_id).to eq(original_instance_id)
+    expect(Unleash.configuration.app_name).to eq(original_app_name)
     scheduled_executor.thread.join
 
-    Unleash.configuration.instance_id = original_instance_id
+    Unleash.configuration.app_name = original_app_name
```
nunogois, line 91: I don't think the property here really mattered (this should work with any property) so I picked `app_name` instead.

sighphyre (commented):
Okaaaaay. I think this looks good but I wanna think a bit about what happens when your application gets forked by gunicorn or something equivalently horrible. I want to say that should be a new instance id

sighphyre (approved):
We manually validated this together, I'm happy it works

PR #181, opened 2024-07-10
fix: move client spec version to source

sighphyre:
Moves the client spec version directly into source code. This was put in the gem spec but that causes requiring the gem to fail under some circumstances


chriswk (approved):
Not seeing anything scary here.

PR #187, opened 2024-07-31
chore: emit warning when overriding a built in strategy

sighphyre:
This emits a std error line when overriding a built in strategy. That's not good behavior. This is just being polite and giving users some breathing room to stop doing that

This is a bit hacky but the intent here is to get people off this behavior so we can merge Yggdrasil (this is effectively the only breaking change that Yggdrasil will cause). Ultimately this code will go away anyway when that's merged so... eh

### DIFF
--- a/lib/unleash/strategies.rb
+++ b/lib/unleash/strategies.rb
@@ -24,8 +24,10 @@ module Unleash
 
     def add(strategy)
       if default_strategy_names.include?(strategy.name)
+        # rubocop:disable Style/StderrPuts
         $stderr.puts "Warning: Overriding built in strategy '#{strategy.name}'. Overriding built in strategies is deprecated and will be \
         removed in a future release."
+        # rubocop:enable Style/StderrPuts
       end
       self.internal_add(strategy)
     end
--- a/lib/unleash/strategies.rb
+++ b/lib/unleash/strategies.rb
@@ -24,7 +24,8 @@ module Unleash
 
     def add(strategy)
       if default_strategy_names.include?(strategy.name)
-        $stderr.puts "Warning: Overriding built in strategy '#{strategy.name}'. Overriding built in strategies is deprecated and will be removed in a future release."
+        $stderr.puts "Warning: Overriding built in strategy '#{strategy.name}'. Overriding built in strategies is deprecated and will be \
+        removed in a future release."
       end
       self.internal_add(strategy)
     end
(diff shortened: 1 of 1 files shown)


lib/unleash/strategies.rb
```diff
@@ -23,7 +23,13 @@ def fetch(name)
     end
 
     def add(strategy)
-      @strategies[strategy.name] = strategy
+      if default_strategy_names.include?(strategy.name)
+        # rubocop:disable Style/StderrPuts
```
sighphyre, line 27: I really, really don't want people to disable this and miss it but I'm not sure what impacts this has on loggers

rarruda, line 27: I personally would be consistent with the rest of the SDK use the logger for this (but use the error log level). And if you strongly believe in throwing this (and other errors) in stderr, configure the default logger to do so.

I am at least a bit fan on being consistent across all of the sdk wrt to logging. 

WDYT?

rarruda: What kind of impact are you thinking about?

sighphyre: What I'm trying to avoid is someone setting a logger config that allows them to ignore or miss this error. I want this to make people sad enough to do something about it today because if you ignore this and upgrade, your toggles will just evaluate in a different way and you'll have no idea why.

Consistency that big of a concern given this won't live very long (I know I'm a bad person).

I'm more worried about not playing nicely with the logger and having this get missed in another way

rarruda: Totally understand the need to be pedantic. Maybe do both?

Hate myself for mentioning, but maybe add something in CAPS too?

If people have log shippers pre-configured, it could get missed if they do something weird with their collections. 

Just my 2c.

sighphyre: Eh. I thought about it. I don't mind that much either way. If you disable errors then your life is going to be hard.

I've moved this to the logger

rarruda: I would have not added it as a warning either. 😛

gastonfournier (approved):
LGTM!

gardleopard (approved):
Looks good to me! Happy that we alert on this!

lib/unleash/strategies.rb
sighphyre: Logger not reading it correctly really, moot point though, moved to a log error now

PR #194, opened 2024-08-29
feat: add segments to new feature toggle

djosino:
## About the changes
I added an optional second parameter Unleash&.segment_cache to the initialization of the method `Unleash::FeatureToggle.new(toggle_as_hash, Unleash&.segment_cache)`.

## Discussion points
Is there another alternative for segments to be considered when querying variant ?

### DIFF
--- a/lib/unleash/client.rb
+++ b/lib/unleash/client.rb
@@ -80,7 +80,7 @@ module Unleash
         return fallback_variant
       end
 
-      toggle = Unleash::FeatureToggle.new(toggle_as_hash)
+      toggle = Unleash::FeatureToggle.new(toggle_as_hash, Unleash&.segment_cache)
       variant = toggle.get_variant(context, fallback_variant)
 
       if variant.nil?


gastonfournier:
Hi @djosino, thanks for the PR, but can you explain what is the purpose or use case of this change? I can see we used the cache elsewhere but I'm not sure what you're trying to accomplish.

djosino:
> Hi @djosino, thanks for the PR, but can you explain what is the purpose or use case of this change? I can see we used the cache elsewhere but I'm not sure what you're trying to accomplish.

Hi @gastonfournier , thanks for your attention.

Yeap, this call is used in the `#is_enabled?` and passing the segments allows the FeatureToggle class to consider the created segments to return the toggle value or variant for that strategy. Without passing the segments, it's only possible to get the return value by configuring constraints in Unleash.

I'll try to send two screenshots to make it clearer.

Unleash configured without segments and with constraints, FeatureToggle without segment cache.
Screenshot_20240830_084607

Unleash configured with segments (same rules), FeatureToggle with and without segment cache.
Screenshot_20240830_085222

djosino:
Hi @gastonfournier.

I don't have much context on how this would apply to the specification project, so it might be quicker if someone with more context could contribute.

In that case, should we keep this PR open or close it?

gastonfournier:
Let's keep it open, we can discuss the spec test with the team and merge this pr next week

gastonfournier:
Hi @djosino I've opened https://github.com/Unleash/client-specification/pull/92 which I verified cover this specific use case (fails without your change and succeeds after your change). I'll wait until tomorrow before merging this so I can align with the team. Just keeping you posted ;)

sighphyre:
@djosino Merged and released under 5.0.7, thank you again!

djosino:
perfect, thanks for the support @sighphyre @gastonfournier

PR #197, opened 2024-09-16
chore: add feature_enabled property to variants

nunogois:
https://linear.app/unleash/issue/2-2638/feature-enabled-support-in-ruby-sdk

Adds a `feature_enabled` property to variants, similar to https://github.com/Unleash/unleash-client-python/pull/285

Also bumps the client spec to the respective version, so the expectation matches.

### DIFF
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -48,11 +48,14 @@ module Unleash
 
       choice = evaluation_result.enabled? ? :yes : :no
       Unleash.toggle_metrics.increment_variant(self.name, choice, variant.name) unless Unleash.configuration.disable_metrics
+
+      variant.feature_enabled = evaluation_result.enabled? || false
+
       variant
     end
 
     def self.disabled_variant
-      Unleash::Variant.new(name: 'disabled', enabled: false)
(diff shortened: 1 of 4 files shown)


lib/unleash/feature_toggle.rb
```diff
@@ -48,11 +48,14 @@ def get_variant(context, fallback_variant = Unleash::FeatureToggle.disabled_vari
 
       choice = evaluation_result.enabled? ? :yes : :no
       Unleash.toggle_metrics.increment_variant(self.name, choice, variant.name) unless Unleash.configuration.disable_metrics
+
+      variant.feature_enabled = evaluation_result.enabled? || false
```
rarruda, line 52: not sure if the `|| false` is needed as `is_enabled?`/`enabled?` methods should really always return a `bool`. (and if it doesn't, that's a pretty big bug somewhere).

nunogois: Thanks, TIL. Addressed in: https://github.com/Unleash/unleash-client-ruby/pull/197/commits/48e4f77cce8aba817833bb34391157f07511026f

rarruda (approved):
Looks sane to me. Just left some bike-shedding comment.

chriswk (approved):
Concur with Renato, this looks sane to me

PR #198, opened 2024-09-16
fix: variant stickiness from strategy

nunogois:
https://linear.app/unleash/issue/2-2644/correctly-resolve-variant-stickiness-from-strategy-in-ruby-sdk

When bumping the client spec in https://github.com/Unleash/unleash-client-ruby/pull/197 to the latest version I noticed we were not being 100% compliant. The reason for it seemed to be the fact that we're not respecting custom strategy stickiness when resolving variants, as running the spec tests multiple times sometimes returned the wrong variant in those tests.

See:
 - https://github.com/Unleash/unleash-client-node/pull/570
 - https://github.com/Unleash/unleash-client-dotnet/pull/212

This PR applies a similar change to the Ruby SDK, bumping the client spec to the latest version.

### DIFF
--- a/lib/unleash/feature_toggle.rb
+++ b/lib/unleash/feature_toggle.rb
@@ -61,17 +61,18 @@ module Unleash
     private
 
     def resolve_variant(context, evaluation_result, group_id)
+      variant_strategy_stickiness = evaluation_result.strategy&.params.to_h['stickiness'] || 'default'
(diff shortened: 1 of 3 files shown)


kwasniew (approved):

gastonfournier (approved):
Nice!

PR #200, opened 2024-09-19
fix: update ruby SDK to increase the possible random numbers used for stickiness id

thomasheartman:
Same fix as done in some other sdks, such as the node one at
https://github.com/Unleash/unleash-client-node/pull/417

Fixes an issue where 1% rollout would not yield any results with
random rollout for certain group ids

### DIFF
--- a/lib/unleash/strategy/flexible_rollout.rb
+++ b/lib/unleash/strategy/flexible_rollout.rb
@@ -40,7 +40,7 @@ module Unleash
       end
 
       def random
-        Random.rand(0..10000)
+        Random.rand(0..10_000)
       end
 
       def resolve_stickiness(stickiness, context)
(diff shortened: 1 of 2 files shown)


ivarconr (approved):

ivarconr:
did you consider a unit test for this? 
https://github.com/Unleash/unleash-client-node/blob/main/src/test/strategy/gradual-rollout-random.test.ts#L10-L33

thomasheartman:
I did not (because there was none in the node sdk where I took it from), but lemme see what I can cook up

thomasheartman:
@ivarconr : that test does not test the fix in the node SDK, which was only applied to the `src/strategy/flexible-rollout-strategy.ts` file. In fact, the `src/strategy/gradual-rollout-random.ts`, which this test presumably tests, still only goes to 101. And interestingly, as a follow-up, if I increase the multiplication to 10k (as we've done in the other file), that test starts failing.

However, the same thing in Ruby does appear to work (fails at 100, works at 10k). I've added the test; will merge when everything goes green unless anyone speaks up.

PR #201, opened 2024-09-19
docs: variant fallback fix

sighphyre:
Variant fallback docs now state that the fallback will be returned when no flag is found on the server, rather than no variant being found on the server


gastonfournier (approved):
Makes sense!

PR #208, opened 2024-09-26
chore: drop ruby 2.6

sighphyre:
Drops support for Ruby 2.6 + JRuby 9.3, which depends on Ruby 2.6.

## Why

Ruby 2.6 has been EOL for over two years and outside normal maintenance for three years. For almost all users this should have been superseded by Ruby 2.7 (also EOL), which is the tip of the 2.x range. Understandably some users won't be able to smoothly move to 3.x, so this is allows us to draw a line in the sand and support the latest 2.7


rarruda (approved):
🎉

chriswk (approved):
Yes, please :)

PR #213, opened 2024-11-13
fix: aarch64 support

sighphyre:
Resolves Yggdrasil core to a version that contains the correct binary for aarch64


gastonfournier (approved):
LG!

PR #219, opened 2024-12-10
Handle empty metrics after 10 min

gastonfournier:
## About the changes
This PR adds a test to validate what happens when metrics are empty for more than 10 minutes. This other PR #218 tested that it was actually failing.

We've also validated that nil bucket returns a 400 from Unleash server side so instead we should always fallback to an empty bucket.

### DIFF
--- a/README.md
+++ b/README.md
@@ -534,7 +534,7 @@ end
 ## Development
 
 After checking out the repo, run `bin/setup` to install dependencies.
-Then, run `rake spec` to run the tests.
+Then, run `bundle exec rake spec` to run the tests.
 You can also run `bin/console` for an interactive prompt that will allow you to experiment.
 
 This SDK is also built against the Unleash Client Specification tests.
@@ -542,7 +542,7 @@ To run the Ruby SDK against this test suite, you'll need to have a copy on your 
 
 `git clone --branch v$(ruby echo_client_spec_version.rb) https://github.com/Unleash/client-specification.git`
 
-After doing this, `rake spec` will also run the client specification tests.
+After doing this, `bundle exec rake spec` will also run the client specification tests.
 
 To install this gem onto your local machine, run `bundle exec rake install`.
 
--- a/lib/unleash/metrics_reporter.rb
+++ b/lib/unleash/metrics_reporter.rb
@@ -32,7 +32,7 @@ module Unleash
       Unleash.logger.debug "post() Report"
 
       bucket = self.generate_report
-      if bucket.nil? && !(Time.now - self.last_time < LONGEST_WITHOUT_A_REPORT) # and last time is less then 10 minutes...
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -534,15 +534,15 @@ end
 ## Development
 
 After checking out the repo, run `bin/setup` to install dependencies.
-Then, run `rake spec` to run the tests.
+Then, run `bundle exec rake spec` to run the tests.
```
gastonfournier, line 537: At least in my machine I needed to add exec and other places in this readme also uses exec so I think this is the safest

lib/unleash/metrics_reporter.rb
```diff
@@ -24,21 +23,19 @@ def generate_report
...
+        'bucket': metrics || {}
       }
     end
 
     def post
       Unleash.logger.debug "post() Report"
-
-      bucket = self.generate_report
-      if bucket.nil? && (Time.now - self.last_time < LONGEST_WITHOUT_A_REPORT) # and last time is less then 10 minutes...
+      report = self.generate_report
```
nunogois, line 32:
```suggestion

      report = self.generate_report
      
```

spec/unleash/metrics_reporter_spec.rb
```diff
@@ -112,6 +113,35 @@
...
+    expect(WebMock).to have_requested(:post, 'http://test-url/client/metrics')
+      .with(
+        body: hash_including(
+          yggdrasilVersion: anything,
+          specVersion: anything,
+          platformName: anything,
+          platformVersion: anything,
+          bucket: {}
+        )
+      )
```
nunogois, line 142: Is this working as expected? I thought the intention would be _not_ to post to server when it's empty.

nunogois: Nvm, we're subtracting by 10 minutes and 1 second, which means that the second part of our condition is not met, so we post the empty metrics. Got it 👍

nunogois (approved):
LGTM

lib/unleash/metrics_reporter.rb
gastonfournier: What's the suggestion here? Adding some space? Just making sure I'm not missing something else

spec/unleash/metrics_reporter_spec.rb
gastonfournier: Indeed, turns out this code path was apparently not tested and didn't work as expected. Although it's an edge case I think it makes sense

lib/unleash/metrics_reporter.rb
nunogois: Yeah, it's no big deal, just keeping the spacing around the statement like before.

PR #221, opened 2024-12-19
Allow symbols as flag names

Bertg:
## About the changes

This PR allows developers to specify flag names as symbols. Ruby developers will often use symbols instead of flag names.

## Discussion points

If this PR is not accepted it would be useful to have some sort of clear error at runtime. Today the lib will raise `TypeError: no implicit conversion of Symbol into String` this is not super useful to the developer.

### DIFF
--- a/lib/unleash/client.rb
+++ b/lib/unleash/client.rb
@@ -36,6 +36,7 @@ module Unleash
     # rubocop:enable Metrics/AbcSize
 
     def is_enabled?(feature, context = nil, default_value_param = false, &fallback_blk)
+      feature = feature.to_s
       Unleash.logger.debug "Unleash::Client.is_enabled? feature: #{feature} with context #{context}"
 
       default_value = if block_given?
@@ -76,6 +77,7 @@ module Unleash
     end
 
     def get_variant(feature, context = Unleash::Context.new, fallback_variant = disabled_variant)
+      feature = feature.to_s
       variant = Unleash.engine.get_variant(feature, context)
 
       if variant.nil?
--- a/spec/unleash/client_spec.rb
+++ b/spec/unleash/client_spec.rb
@@ -541,6 +541,78 @@ RSpec.describe Unleash::Client do
     expect(unleash_client.disabled?('any_feature', {}, false)).to eq(false)
   end
 
+  it "should accept symbols as flag names", focus: true do
+    WebMock.stub_request(:post, "http://test-url/client/register")
+      .with(
+        headers: {
+          'Accept' => '*/*',
+          'Content-Type' => 'application/json',
+          'Accept-Encoding' => 'gzip;q=1.0,deflate;q=0.6,identity;q=0.3',
+          'Unleash-Appname' => 'my-test-app',
+          'Unleash-Instanceid' => 'rspec/test',
+          'User-Agent' => "UnleashClientRuby/#{Unleash::VERSION} #{RUBY_ENGINE}/#{RUBY_VERSION} [#{RUBY_PLATFORM}]",
+          'X-Api-Key' => '123'
+        }
+      )
+      .to_return(status: 200, body: "", headers: {})
+
+    features_response_body = '{
+      "version": 1,
+      "features": [
+        "name": "toggleNameAgain",
+        "enabled": true,
+        "strategies": [{ "name": "default" }],
+        "variants": [
+          {
+            "name": "a",
+            "weight": 50,
+            "payload": {
+              "type": "string",
+              "value": ""
+            }
+          },
+          {
+            "name": "b",
+            "weight": 50,
+            "payload": {
+              "type": "string",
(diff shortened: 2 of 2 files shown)


Bertg:
@gastonfournier one possible change would be to move this to `Unleash.engine` / `yggdrasil-engine/ but I'm not sure how that one works.

gastonfournier:
> @gastonfournier one possible change would be to move this to `Unleash.engine` / `yggdrasil-engine/ but I'm not sure how that one works.

I think it's better how you did it, it's better to convert feature names to string in the edge of the API layer, so internally we only worry about feature as strings. That's also what the other SDKs do, the strongly typed ones only accept strings but in the case of Ruby, calling to_s on the feature seems correct.

gastonfournier:
An alternative would be to validate that the feature is a string and thrown an error, but that would result in a runtime error with the only difference that the error could be more meaningful. The point being that Unleash API only accepts strings, but this feels like a gray area for languages like ruby where you can't force a type constraint at compile time...

gastonfournier:
We've talked about it with @sighphyre and we want to also get insights from @rarruda before making a final decision. Just keeping you posted.

sighphyre:
Hey @Bertg, so  I actually quite like this. I think it's consistent with the way we do context fields and generally with most Ruby development.

I have two small concerns:

1) I think blindly calling `to_s` on arbitrary objects is going to make life really hard when someone eventually passes in something they shouldn't here. I'd like to bound this to only strings and symbols (I realize what we did in the past was also awful but this is a good chance to make this sane)
2) I would like to get @rarruda's inputs on this, I believe he had some concerns

Other than that I think this is worth doing

Bertg:
Ik, I can adjust the PR. But what should we do if we receive something that's is neither a string or symbol?

I have a few suggestions:

* raise RuntimeError, simple clean
* allow it to happen, maybe warn
* Invoke a specific method. Something like `.to_feature_name` or `.to_feature`. (I've seen this in other libraries, but I think this is probably out of scope for this). 

I would opt for `warn` now, mark it is deprecated behaviour. Allow the user to switch to a strict mode (`raise`), which would become the default in a new major version. 

But your call, and I'll adjust the PR

sighphyre:
> OK, I can adjust the PR. But what should we do if we receive something that's is neither a string or symbol?
> 
> I have a few suggestions:
> 
>     * raise RuntimeError, simple clean
> 
>     * allow it to happen, maybe warn
> 
>     * Invoke a specific method. Something like `.to_feature_name` or `.to_feature`. (I've seen this in other libraries, but I think this is probably out of scope for this).
> 
> 
> I would opt for `warn` now, mark it is deprecated behaviour. Allow the user to switch to a strict mode (`raise`), which would become the default in a new major version.
> 
> But your call, and I'll adjust the PR

I'm tempted to say "let it happen, write something to the error log rather than warn". It's a pretty serious issue in that your flags are _definitely_ not evaluating the way you think they should be but the philisophy of all our SDKs is to default to false rather than fail when and where they can (I'm pretty sure passing goop will result in a false here)

rarruda:
I feel extremely conflicted. I agree a lot with the statement:

> Ruby developers will often use symbols instead of flag names.

That is definitely best practice.

But in my opinion the following error is descriptive:
> `TypeError: no implicit conversion of Symbol into String` this is not super useful to the developer.

But, again, the SDK probably shouldn't be throwing exceptions, just logging a WARN/ERROR and returning false, as @sighphyre mentioned.

Keeping accepting just strings, makes a ton of sense from the unleash API point of view. Adding an implicit conversion in the SDK feels a bit dirty though. I'd like to let this simmer in my brain a bit more before taking a side.

----

Side note: I feel like a `PermissiveUnleashClient`, or maybe having a config setting that allows a more permissive could be a way around this, but i also feel a bit uneasy about it. This is a non-tested version of the idea/concept:
```
module Unleash::Extensions::AllowSymbols
  def is_enabled?(feature, context = nil, default_value_param = false, &fallback_blk)
    super(feature.to_s, context, default_value_param, &fallback_blk)
  end

  def get_variant(feature, context = Unleash::Context.new, fallback_variant = disabled_variant)
    super(feature.to_s, context, fallback_variant)
  end
end

module Unleash
  class Client
    prepend Unleash::Extensions::AllowSymbols if Unleash.configuration.allow_symbols
  end
end
```

But I can say for sure, is that it makes a ton of sense for us to document a bit more explicitly the expected types. Maybe the nudge i need to bring back to life the  work in the `feat/add_rbs` branch.

Bertg:
Something I noticed today... In the [Variant class](https://github.com/Unleash/unleash-client-ruby/blob/a92677155b06208267f1e4d706d04c07182641b6/lib/unleash/variant.rb#L8) the library is already dealing with the Symbol/String ambiguity by allowing both.

Just an FYI.

PR #223, opened 2025-01-08
chore: add ruby 3.4 support in gh actions

rarruda:
## About the changes
Add github actions test also in the latest ruby release.


rarruda:
Oops! Looks like some changes are needed to make all tests pass as green in `3.4`. Will take a look when i get a chance.

sighphyre (approved):
Awesome, that looks sane to me

PR #224, opened 2025-01-20
feat: client identification headers

kwasniew:
## About the changes

This PR adds standardised client identification headers to the feature and metrics calls that the client makes to Unleash. The headers are:
- `x-unleash-appname`: the name of the application that is using the client. `UNLEASH-APPNAME` will be deleted in another PR (expand/contract pattern)
- `x-unleash-connection-id`: an internal unique identifier for the current instance of the client generated by the built-in crypto lib
- `x-unleash-sdk`: sdk information in the format `unleash-ruby@<version>` following reference `unleash-node@<version>` naming pattern

All the headers are intended for the Unleash team so clients should not be affected.
The main use cases we have are:
* capacity planning by knowing the number of unique connections made to the backend API
* debugging misconfigured SDKs sending more traffic than expected

Reference implementation: https://github.com/Unleash/unleash-client-node/pull/690

### DIFF
--- a/lib/unleash/configuration.rb
+++ b/lib/unleash/configuration.rb
@@ -103,8 +103,7 @@ module Unleash
       self.strategies       = Unleash::Strategies.new
 
       self.custom_http_headers = {}
-
-      @connection_id        = SecureRandom.uuid
+      @connection_id = SecureRandom.uuid
     end
 
     def initialize_default_logger
--- a/lib/unleash/configuration.rb
+++ b/lib/unleash/configuration.rb
@@ -54,6 +54,9 @@ module Unleash
         'User-Agent' => "UnleashClientRuby/#{Unleash::VERSION} #{RUBY_ENGINE}/#{RUBY_VERSION} [#{RUBY_PLATFORM}]",
         'UNLEASH-INSTANCEID' => self.instance_id,
         'UNLEASH-APPNAME' => self.app_name,
+        'X-UNLEASH-APPNAME' => self.app_name,
+        'X-UNLEASH-CONNECTION-ID' => @connection_id,
+        'X-UNLEASH-SDK' => "unleash-ruby@#{Unleash::VERSION}",
         'Unleash-Client-Spec' => CLIENT_SPECIFICATION_VERSION
       }.merge!(generate_custom_http_headers)
     end
@@ -87,6 +90,7 @@ module Unleash
       self.environment      = 'default'
       self.url              = nil
       self.instance_id      = SecureRandom.uuid
+      @connection_id        = SecureRandom.uuid
       self.project_name     = nil
       self.disable_client   = false
       self.disable_metrics  = false
--- a/lib/unleash/configuration.rb
+++ b/lib/unleash/configuration.rb
@@ -90,7 +90,6 @@ module Unleash
       self.environment      = 'default'
       self.url              = nil
       self.instance_id      = SecureRandom.uuid
-      @connection_id        = SecureRandom.uuid
       self.project_name     = nil
       self.disable_client   = false
       self.disable_metrics  = false
@@ -104,6 +103,8 @@ module Unleash
       self.strategies       = Unleash::Strategies.new
 
       self.custom_http_headers = {}
+
+      @connection_id        = SecureRandom.uuid
     end
 
     def initialize_default_logger
--- a/spec/unleash/client_spec.rb
+++ b/spec/unleash/client_spec.rb
@@ -1,3 +1,5 @@
+require "securerandom"
+
 RSpec.describe Unleash::Client do
   after do
     WebMock.reset!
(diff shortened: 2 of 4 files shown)


lib/unleash/configuration.rb
```diff
@@ -54,6 +54,9 @@ def http_headers
         'User-Agent' => "UnleashClientRuby/#{Unleash::VERSION} #{RUBY_ENGINE}/#{RUBY_VERSION} [#{RUBY_PLATFORM}]",
         'UNLEASH-INSTANCEID' => self.instance_id,
         'UNLEASH-APPNAME' => self.app_name,
+        'X-UNLEASH-APPNAME' => self.app_name,
+        'X-UNLEASH-CONNECTION-ID' => @connection_id,
```
kwasniew, line 58: we want this one to be private internal detail of each SDK

lib/unleash/configuration.rb
```diff
@@ -54,6 +54,9 @@ def http_headers
...
+        'X-UNLEASH-SDK' => "unleash-ruby@#{Unleash::VERSION}",
```
kwasniew, line 59: common naming convention across SDKs

lib/unleash/configuration.rb
```diff
@@ -54,6 +54,9 @@ def http_headers
         'User-Agent' => "UnleashClientRuby/#{Unleash::VERSION} #{RUBY_ENGINE}/#{RUBY_VERSION} [#{RUBY_PLATFORM}]",
         'UNLEASH-INSTANCEID' => self.instance_id,
         'UNLEASH-APPNAME' => self.app_name,
+        'X-UNLEASH-APPNAME' => self.app_name,
```
kwasniew, line 57: this will replace `UNLEASH-APPNAME` over time

spec/unleash/client_spec.rb
```diff
@@ -1,17 +1,23 @@
...
+
 RSpec.describe Unleash::Client do
   after do
     WebMock.reset!
     File.delete(Unleash.configuration.backup_file) if File.exist?(Unleash.configuration.backup_file)
   end
 
   it "Uses custom http headers when initializing client" do
+    fixed_uuid = "123e4567-e89b-12d3-a456-426614174000"
+    allow(SecureRandom).to receive(:uuid).and_return(fixed_uuid)
```
kwasniew, line 11: since we don't expose connection id we need to mock the generation

lib/unleash/configuration.rb
```diff
@@ -87,6 +90,7 @@ def set_defaults
       self.environment      = 'default'
       self.url              = nil
       self.instance_id      = SecureRandom.uuid
+      @connection_id        = SecureRandom.uuid
```
sighphyre, line 93: This is the dumbest comment I've ever left on a PR ever but the change in syntax here is a bit jarring. Any chance we can move this line either to the top of the bottom of this function?

Honestly I won't even be offended if you just ignore me

sighphyre (approved):
Yep, seems reasonable to me!

lib/unleash/configuration.rb
kwasniew: It actually makes perfect sense to me. Had the same thought :) Moving it to the bottom

sighphyre: Thank you! I appreciate you haha

rarruda (commented):
Mostly bike shedding, but since you ask:

1. Any reasons why you use uppercase for the headers here in the Ruby SDK but lowercase in the node SDK? (IIRC the mocks ignore the case in the headers, but that doesn't mean that we should have them different across different SDKs, if the point is to standardize)
2. Any links to any more formal docs/specs? I think a link to the PR in the node SDK is ok too, but would be even better if there was a link to something a tiny bit more formal.
3. In an ideal world the commit message would include:
  - the migration path between the headers (which header is going to be replacing which),
  - the rationale ("standardization across SDKs" should suffice, but "client identification headers" is a bit on the opaque side for my taste). 
  - Ideally also add a link/hook to point 2. (either which commit in the node SDK is used as a base, or a link to the docs/spec).

But the code itself is quite sane and i enjoy nit-picking. So if you feel like merging as is, I don't frown for a moment either.

kwasniew:
@rarruda Thanks so much for taking the time to review this PR. 
To answer your comments:
1. we keep the existing convention in this SDK (`UNLEASH-INSTANCEID` and `UNLEASH-APPNAME`). Since HTTP headers are case insensitive we're safe here no only in tests but also in the production setup. I noticed that some SDKs like Java use upper case while some others like Node use lower-case. If we decide to unify the casing I'd like to do it in another PR. 
2. Since this is internal change not affecting clients of the SDK (only Unleash core team and SDK maintainers) I linked to the reference implementation only. However I'm enriching the PR description based on feedback gathered so far. At some point we may introduce executable spec for the HTTP contract as we have for the flag evaluations and that would be the ideal case. 
3. I expanded the PR description with the use case description and the explanation that users of the SDK are not affected by this change.

PR #225, opened 2025-01-21
chore: bump version to 6.1.0

kwasniew:
## About the changes

Closes #


sighphyre (approved):
Personally I think this should be a patch but that's not a blocking opinion so LGTM

PR #228, opened 2025-01-29
Chore/update ygg engine

FredrikOseberg:
This PR updates yggdrasil engine to 1.0.2 and prepares for an alpha release of the Ruby SDK

### DIFF
--- a/lib/unleash/toggle_fetcher.rb
+++ b/lib/unleash/toggle_fetcher.rb
@@ -81,6 +81,8 @@ module Unleash
 
       # notify all threads waiting for this resource to no longer wait
       self.toggle_resource.broadcast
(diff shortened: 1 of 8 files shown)


lib/unleash/toggle_fetcher.rb
```diff
@@ -82,6 +82,8 @@ def update_engine_state!(toggle_data)
 
       # notify all threads waiting for this resource to no longer wait
       self.toggle_resource.broadcast
+    rescue StandardError => e
```
sighphyre, line 85: This causes the SDK to eat this error without logging anything so this change gives us some insight into yggdrasil failures

sighphyre (approved):
LGTM

PR #238, opened 2025-04-09
Relax logger dependency

Drowze:
## About the changes
Relax `logger` gem dependency, so unleash can be used with logger >= 1.6.5.

This should allow using unleash gem with [logger 1.7.0](https://github.com/ruby/logger/releases/tag/v1.7.0) (which was released in 26/March).

Tests seem to pass locally.


rarruda (approved):
I doubt that logger breaks it's API in a minor version change. If all tests pass as expected, then there is no reason to not relax the dependency version.

PR #240, opened 2025-04-14
chore: update rubocop

sighphyre:
CI uses latest and dev is lagging, which gives a very unpleasant surprise when you land your PR on Github and suddenly Rubocop has _**opinions**_

This is just an update to Rubocop 1.75 + apply Rubocop fixes

### DIFF
--- a/lib/unleash/metrics_reporter.rb
+++ b/lib/unleash/metrics_reporter.rb
@@ -14,7 +14,7 @@ module Unleash
     end
 
     def generate_report
-      metrics = Unleash&.engine&.get_metrics()
+      metrics = Unleash.engine&.get_metrics
 
       {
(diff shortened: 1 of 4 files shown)


lib/unleash/metrics_reporter.rb
```diff
@@ -14,7 +14,7 @@ def initialize
     end
 
     def generate_report
-      metrics = Unleash&.engine&.get_metrics()
+      metrics = Unleash.engine&.get_metrics
```
nunogois, line 17: Just confirming: Is this really the same thing in Ruby?

sighphyre: Yeah, it's weird I know but it is the same thing

PR #248, opened 2025-08-05
feat: streaming support

kwasniew:
## About the changes

Adding SSE streaming client support in this SDK similar to Node SDK.

* When starting the SDK we can set the experimental mode to be streaming instead of default polling mode
* I manually tested that disconnects are handled gracefully (this should be also tested in the SSE client itself)
* Added tests showing how streaming works from this SDK perspective
* Under the hood streaming events are handled by the yggdrasil engine so we don't need to test all the details as in the Node SDK
* JRuby doesn't have support for one of the SSE client transitive dependencies so it falls back to polling
* Also we don't store state to a backup file for now in streaming mode. It will be added once yggdrasil can expose full state to the SDKs

### DIFF
--- a/lib/unleash/configuration.rb
+++ b/lib/unleash/configuration.rb
@@ -97,13 +97,8 @@ module Unleash
     end
 
     def streaming_mode?
-      is_streaming_configured = self.experimental_mode.is_a?(Hash) && self.experimental_mode[:type] == 'streaming'
-
-      if is_streaming_configured && RUBY_ENGINE == 'jruby'
-        raise "Streaming mode is not supported on JRuby. Please use polling mode instead."
-      end
-
-      is_streaming_configured
+      validate_streaming_support! if streaming_configured?
+      streaming_configured?
     end
 
     def polling_with_delta?
@@ -168,6 +163,16 @@ module Unleash
       __send__("#{opt}=", val)
     rescue NoMethodError
       raise ArgumentError, "unknown configuration parameter '#{val}'"
+    end
+
+    def streaming_configured?
+      self.experimental_mode.is_a?(Hash) && self.experimental_mode[:type] == 'streaming'
+    end
+
+    def validate_streaming_support!
+      return unless RUBY_ENGINE == 'jruby'
+
+      raise "Streaming mode is not supported on JRuby. Please use polling mode instead or switch to MRI/CRuby."
     end
   end
 end
--- a/lib/unleash/streaming_client_executor.rb
+++ b/lib/unleash/streaming_client_executor.rb
@@ -19,21 +19,23 @@ module Unleash
     def start
       return if self.running || Unleash.configuration.disable_client
 
-      Unleash.logger.debug "Starting streaming executor from URL: #{Unleash.configuration.fetch_toggles_uri}"
+      Unleash.logger.debug "Streaming client #{self.name} starting connection to: #{Unleash.configuration.fetch_toggles_uri}"
 
       self.event_source = create_event_source
       setup_event_handlers
 
       self.running = true
+      Unleash.logger.debug "Streaming client #{self.name} connection established"
     end
 
     def stop
       return unless self.running
 
-      Unleash.logger.info "Stopping streaming executor"
+      Unleash.logger.debug "Streaming client #{self.name} stopping connection"
       self.running = false
       self.event_source&.close
       self.event_source = nil
+      Unleash.logger.debug "Streaming client #{self.name} connection closed"
     end
 
     alias exit stop
@@ -69,14 +71,15 @@ module Unleash
       end
 
       self.event_source.on_error do |error|
-        Unleash.logger.warn "Streaming error: #{error}"
+        Unleash.logger.warn "Streaming client #{self.name} error: #{error}"
       end
     end
 
     def handle_event(event)
       self.event_processor.process_event(event)
     rescue StandardError => e
-      Unleash.logger.error "Error in streaming executor event handling: #{e.message}"
+      Unleash.logger.error "Streaming client #{self.name} threw exception #{e.class}: '#{e}'"
+      Unleash.logger.debug "stacktrace: #{e.backtrace}"
     end
   end
 end
--- a/lib/unleash/streaming_event_processor.rb
+++ b/lib/unleash/streaming_event_processor.rb
@@ -21,7 +21,8 @@ module Unleash
         Unleash.logger.debug "Received unknown event type: #{event.type}"
       end
     rescue StandardError => e
-      Unleash.logger.error "Error handling streaming event: #{e.message}"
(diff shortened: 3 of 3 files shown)


examples/simple.rb
```diff
@@ -18,17 +18,16 @@
 # or:
 
 @unleash = Unleash::Client.new(
-  url: 'https://unleash.herokuapp.com/api',
-  custom_http_headers: { 'Authorization': '943ca9171e2c884c545c5d82417a655fb77cec970cc3b78a8ff87f4406b495d0' },
+  url: 'https://app.unleash-hosted.com/demo/api',
```
kwasniew, line 21: re-using sdk-examples setup https://github.com/Unleash/unleash-sdk-examples/blob/main/Ruby/.env.example as heroku is deprecated

lib/unleash/client.rb
```diff
@@ -21,7 +23,8 @@ def initialize(*opts)
       Unleash.engine = YggdrasilEngine.new
       Unleash.engine.register_custom_strategies(Unleash.configuration.strategies.custom_strategies)
 
-      Unleash.toggle_fetcher = Unleash::ToggleFetcher.new Unleash.engine
+      Unleash.toggle_fetcher = Unleash::ToggleFetcher.new Unleash.engine unless Unleash.configuration.streaming_mode?
```
kwasniew, line 26: fetcher makes initial HTTP call on instance creation. We don't want that for how with streaming to avoid cross locks between streaming client and polling client

lib/unleash/configuration.rb
```diff
@@ -22,7 +22,8 @@ class Configuration
       :log_level,
       :bootstrap_config,
       :strategies,
-      :use_delta_api
+      :use_delta_api,
```
kwasniew, line 25: we should deprecate this property as we have a standardized experimental mode in Node SDK that is either {type: 'streaming'} or {type: 'polling', format: 'delta'} or {type: 'polling', format: 'full'}

lib/unleash/streaming_client.rb
```diff
@@ -0,0 +1,93 @@
...
+      Unleash.logger.debug "Processing initial hydration data"
+      handle_updated_event(event)
+    end
+
+    def handle_updated_event(event)
+      self.mutex.synchronize do
+        self.toggle_engine.take_state(event.data)
+      end
+
+      # TODO: update backup file
```
kwasniew, line 86: state engine needs to expose current state so we can save it . Both streaming and polling with delta will need it

examples/streaming.rb
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env ruby
+
+require 'unleash'
+require 'unleash/context'
+
+puts ">> START streaming.rb"
+
+@unleash = Unleash::Client.new(
+  url: 'https://app.unleash-hosted.com/demo/api',
```
kwasniew, line 9: demo doesn't have streaming enabled so this needs to be changed to enterprise URL

spec/unleash/client_spec.rb
```diff
@@ -761,4 +761,89 @@ def enabled?(_params, context)
       unleash_client.is_enabled?('featureX', Unleash::Context.new({}))
     ).to be false
   end
+
+  unless RUBY_ENGINE == 'jruby'
```
kwasniew, line 765: run those tests for non jruby platforms

lib/unleash/streaming_client.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    def initialize(toggle_engine)
+      self.toggle_engine = toggle_engine
+      self.event_source = nil
+      self.running = false
+      self.mutex = Mutex.new
+    end
+
+    def start
+      self.mutex.synchronize do
```
kwasniew, line 15: similar synchronization pattern to toggle_fetcher so that we guard running property

lib/unleash/configuration.rb
```diff
@@ -93,6 +96,18 @@ def use_bootstrap?
       self.bootstrap_config&.valid?
     end
 
+    def streaming_mode?
+      return false if RUBY_ENGINE == 'jruby'
```
sighphyre, line 100: I think we should consider raising an exception here or logging an error. I would be very annoyed if an SDK I used didn't do what I told it and quietly did something different

lib/unleash/streaming_client.rb
sighphyre: Why do we need to guard running? That should only ever be a problem if two threads are calling start and stop on the same instance in quick succession, surely?

lib/unleash/client.rb
```diff
@@ -2,14 +2,16 @@
 require 'unleash/toggle_fetcher'
```
sighphyre, line 2: I feel like there's a missing somewhere here and I think that's related to the difference between the toggle fetcher internals and LD's event source lib. The toggle fetching code splits the logic for fetching toggles and the concurrency responsibility into two classes, whereas with the event source lib it's a single entity. This means we can't treat them like interchangable ducks and I really think we should be able to do that. It feels very wrong to have a `:streaming_client` and `:fetcher_scheduled_executor` exposed as accesors when we can only ever have one of these and they do the same job in different ways. It's leading to a bunch of places in this PR where we're checking which we and making local decisions. If we created something like a `fetch_client` that composed of both the `toggle_fetcher` and `scheduled_executor` which exposes start and stop methods we could make a lot of that go away

lib/unleash/util/event_source_wrapper.rb
```diff
@@ -0,0 +1,17 @@
+module Unleash
+  module Util
+    module EventSourceWrapper
+      def self.client
```
sighphyre, line 4: Hmm, I'm a bit tempted to fail hard here rather than fall back to polling. I'm kinda okay with errors that happen on startup and I think it's better to force folks to set their stuff up correctly. We can always relax it later but if we do this now, I doubt we can enforce it later

sighphyre (commented):
Cool! Come grab me next week and we can beat Yggdrasil into shape so you can finish this off!

lib/unleash/configuration.rb
kwasniew: fair point

lib/unleash/client.rb
kwasniew: You comment was spot on. Thank you for that. I managed to create streaming_client_executor and streaming_event_processor. streaming_client_executor can play a role of a fetcher_scheduled_executor (kept the same field name in attr jus tin case someone depends on it). Please let me know if you like it more now

lib/unleash/streaming_client.rb
kwasniew: No mutex needed in the scheduler for SSEs - lifecycle methods called from single thread. But in the event processor event processing happens from EventSource background threads that could potentially be concurrent so keeping it there

sighphyre (approved):
Looking awesome. Gonna be great to get this one done!

PR #255, opened 2025-08-28
feat: read backup file and bootstrap for streaming

kwasniew:
## About the changes

* extracted backup file reading from a polling toggle fetcher
* reading backup file on startup in streaming so even on unsuccessful connection we get some backup
* also added bootstrap reading in streaming and if the bootstrap fails we fall back to backup file
* deliberately keeping backup and bootstrap duplicated between polling and streaming to avoid premature abstraction. it was enough to extract the shared backup file reader
* added tests describing missing behavior (tests inspired by toggle_fetcher_spec.rb)

## Discussion points

We may consider reading backup file outside of the fetcher/streamer if we change current architecture a bit

### DIFF
--- a/lib/unleash/backup_file_reader.rb
+++ b/lib/unleash/backup_file_reader.rb
@@ -1,5 +1,4 @@
 require 'unleash/configuration'
-require 'json'
 
 module Unleash
   class BackupFileReader
@@ -13,11 +12,6 @@ module Unleash
     rescue IOError => e
       # :nocov:
       Unleash.logger.error "Unable to read the backup_file: #{e}"
-      # :nocov:
-      nil
-    rescue JSON::ParserError => e
-      # :nocov:
-      Unleash.logger.error "Unable to parse JSON from existing backup_file: #{e}"
       # :nocov:
       nil
     rescue StandardError => e
--- a/lib/unleash/streaming_client_executor.rb
+++ b/lib/unleash/streaming_client_executor.rb
@@ -21,7 +21,7 @@ module Unleash
           read_backup_file!(engine)
         end
       rescue StandardError => e
-        # fail back to reading the backup file
+        # fall back to reading the backup file
         Unleash.logger.warn "StreamingClientExecutor was unable to initialize, attempting to read from backup file."
         Unleash.logger.debug "Exception Caught: #{e}"
         read_backup_file!(engine)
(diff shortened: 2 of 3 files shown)


lib/unleash/backup_file_reader.rb
```diff
@@ -0,0 +1,30 @@
+require 'unleash/configuration'
+require 'json'
+
+module Unleash
+  class BackupFileReader
```
kwasniew, line 5: this is extracted from toggle_fetcher

lib/unleash/streaming_client_executor.rb
```diff
@@ -10,6 +12,20 @@ def initialize(name, engine)
       self.event_source = nil
       self.event_processor = Unleash::StreamingEventProcessor.new(engine)
       self.running = false
+
+      begin
```
kwasniew, line 16: this was missing

lib/unleash/backup_file_reader.rb
```diff
@@ -0,0 +1,30 @@
...
+      backup_file = Unleash.configuration.backup_file
+      return nil unless File.exist?(backup_file)
+
+      File.read(backup_file)
+    rescue IOError => e
+      # :nocov:
+      Unleash.logger.error "Unable to read the backup_file: #{e}"
+      # :nocov:
+      nil
+    rescue JSON::ParserError => e
```
sighphyre, line 18: I actually don't think this case can happen since we no longer parse on read. I realise this is a hangover from the old implementation (my bad, should have dropped this when I ported to Ygg). If we don't remove it in this PR, no biggie, I'll do it in a future one

lib/unleash/streaming_client_executor.rb
```diff
@@ -10,6 +12,20 @@ def initialize(name, engine)
...
+        # if bootstrap configuration is available, initialize. Otherwise read backup file
+        if Unleash.configuration.use_bootstrap?
+          bootstrap(engine)
+        else
+          read_backup_file!(engine)
+        end
+      rescue StandardError => e
+        # fail back to reading the backup file
```
sighphyre, line 24: Also a hangover from the old code haha, but may as well fix it while we're here
```suggestion
        # fall back to reading the backup file
```

sighphyre (approved):
Nice, I like the harvested backup file reader!

PR #265, opened 2026-05-29
chore: drop Ruby 2.7 and 3.0

sighphyre:
These are now going on 3 years EOL and there are two EOL versions that we support after these two. I think that's reasonable legacy support


chriswk (approved):
Yes please 👍

rarruda (approved):

PR #266, opened 2026-07-07
chore: add a publish pipeline

sighphyre:


rarruda:
This looks awesome! 

The only thing I could think to bikeshed about is it checking if the changelog was already updated from before. (And either have a warning, or automatically populate it? 🤷 or not?)

I assume it is somewhat standardized wrt the other SDKs.

Again awesome work!

sighphyre:
> The only thing I could think to bikeshed about is it checking if the changelog was already updated from before. (And either have a warning, or automatically populate it? 🤷 or not?)

Oh good point I'll take a look before I merge

> I assume it is somewhat standardized wrt the other SDKs.

Trying *sigh* lotta work but I am trying to get them standardized haha

PR #267, opened 2026-07-17
feat: update client with sdk_flavor_fields

irstavr:


lib/unleash/configuration.rb
```diff
@@ -40,6 +42,13 @@ def metrics_interval_in_millis
       self.metrics_interval * 1_000
     end
 
+    def sdk_flavor_fields
+      fields = {}
+      fields[:sdkFlavor] = self.sdk_flavor if self.sdk_flavor
+      fields[:sdkFlavorVersion] = self.sdk_flavor_version if self.sdk_flavor_version
+      fields
+    end
```
sighphyre, line 50: I think this is fine. We can make it a little more terse and leverage Ruby's magic to strip away the nils without having to worry about them

```suggestion
  def sdk_flavor_fields
    {
      sdkFlavor: self.sdk_flavor,
      sdkFlavorVersion: self.sdk_flavor_version
    }.compact
  end
```

sighphyre (approved):
Sure I think this is reasonable. You gotta fight with Rubocop a little here though

